PR #31, opened 2017-12-21
COMP: allow examples to build when ITK_LEGACY_REMOVE is enabled

dzenanz:


dzenanz:
Wait, I should have pushed this to Gerrit.

PR #32, opened 2017-12-21
BUG: Revert updating ImageReadExtractFilterInsertWrite test baseline

fbudin69500:
Change-Id: I3ca570b12ca8418f11444bcfb1251d70aeefac72


fbudin69500:
It does on my machine, both on Windows and Linux. Caveat: I didn't start from a clean build.

fbudin69500:
Actually after a clean build it failed.... I think the first problem on my computer was due to a dirty build.

PR #33, opened 2017-12-24
BUG: Remove FlipImageFilter

thewtex:
This conflicts with the FlipImageFilter executable from ITK/Examples when the
sphinx examples are built as a remote module. Not needed -- we already have
Flip An Image Over Specified Axes

Change-Id: Iad217d722d71d5aba86cd39cb1fa4dfed1e50367


hjmjohnson:
@thewtex Did you verify that both BUILD_EXAMPLES and this module can now be configured? 

Thanks,
Hans

thewtex:
@hjmjohnson I'll follow-up with other conflicts with `BUILD_EXAMPLES` in additional PRs.

PR #35, opened 2017-12-26
Address BUILD_DOCUMENTATION sphinx warnings

thewtex:


thewtex:
@jhlegarreta I took quite a few passes at the `AdaptiveHistogram` issue, but we finally in a good place :-)

With the warnings cleaned up, it will be easier to prevent issues from slipping in.

PR #37, opened 2018-01-14
ENH: Require CMake minimum version to be 3.8.2.

jhlegarreta:
Require CMake minimum version to be 3.8.2 following ITKv5 requiring
C++11:
https://discourse.itk.org/t/minimum-cmake-version-update/585

Change-Id: I491763a77f8434430c8e479e4383dfd92583ede8

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.8.2)
 foreach(p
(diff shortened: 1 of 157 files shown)


jhlegarreta:
@hjmjohnson @thewtex @dzenanz @blowekamp @fbudin69500 Waiting for the discussion in the mentioned thread to reach consensus to push a new patch set with `3.9.5` if required.

thewtex:
This was addressed in #38

PR #46, opened 2018-02-26
Registration readability

thewtex:


src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
```diff
@@ -27,95 +27,70 @@
...
-    FixedImageType, itk.D]
-MovingLinearInterpolatorType = itk.LinearInterpolateImageFunction[
-    MovingImageType, itk.D]
 
-fixedInterpolator = FixedLinearInterpolatorType.New()
-movingInterpolator = MovingLinearInterpolatorType.New()
-
-metric.SetFixedInterpolator(fixedInterpolator)
-metric.SetMovingInterpolator(movingInterpolator)
+TransformType = itk.TranslationTransform[itk.D, Dimension]
```
glemaitre: Could we deduce `Dimension` implicitly?

glemaitre: Can we use something else than `itk.D`. It seems that `itk.F` does not work. So somehow, it should not require anything from the user, isn't it?

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
```diff
@@ -27,95 +27,70 @@
...
+optimizer = itk.RegularStepGradientDescentOptimizerv4[itk.D].New(
+        LearningRate=4,
+        MinimumStepLength=0.001,
+        RelaxationFactor=0.5,
+        NumberOfIterations=200)
 
-fixedImageReader.SetFileName(fixedImage)
-movingImageReader.SetFileName(movingImage)
+metric = itk.MeanSquaresImageToImageMetricv4[
+    FixedImageType, MovingImageType].New()
```
glemaitre: This is one of the case that it seems unnatural. I would think that the image type could be deduced from the input image. It was the reason why I thought that it could be possible to instantiate the metric within the `ImageRegistrationMethod`.

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
```diff
@@ -27,95 +27,70 @@
...
-metric.SetFixedInterpolator(fixedInterpolator)
-metric.SetMovingInterpolator(movingInterpolator)
+TransformType = itk.TranslationTransform[itk.D, Dimension]
+initialTransform = TransformType.New()
 
-FixedImageReaderType = itk.ImageFileReader[FixedImageType]
-MovingImageReaderType = itk.ImageFileReader[MovingImageType]
-fixedImageReader = FixedImageReaderType.New()
-movingImageReader = MovingImageReaderType.New()
+optimizer = itk.RegularStepGradientDescentOptimizerv4[itk.D].New(
```
glemaitre: `itk.D` could be deduced no? Would it make sense to have a `double` in transform and only `float` in the optimizer?

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
```diff
@@ -27,95 +27,70 @@
...
+        LearningRate=4,
```
glemaitre: I assume that the CapWords convention is coming from the C++ and there is no way to change it ;)
I think that PEP8 would prefer something like `learning_rate`.

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
thewtex: Yes we could customize the instantiation here:

  https://github.com/InsightSoftwareConsortium/ITK/blob/e41790229b92fd98865037153ac631e68ac0e9fa/Wrapping/Generators/Python/itkTemplate.py#L382-L383

so that `itk.TranslationTransform.New(image)` picks the the right `Dimension`.

thewtex: Yes, we could just use `itk.D` per above.

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
thewtex: We could also add the logic here:

  https://github.com/InsightSoftwareConsortium/ITK/blob/e41790229b92fd98865037153ac631e68ac0e9fa/Wrapping/Generators/Python/itkTemplate.py#L382-L383

for images so that `itk.MeanSquaresImageToImageMetricv4.New(FixedImage=fixedImage)` returns the right metric.

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
thewtex: If we omit the template type, it just picks the first one, which is `itk.D`, which is fine in this case.  I will add another commit in this branch.

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
thewtex: We can do it! :-)

Please review this patch:

  http://review.source.kitware.com/#/c/23213/

thewtex:
@glemaitre Thanks for the review and thoughts!

> 
> I have an additional question. instantiating as:
>
> instance = itk.ClassITK[arg1, arg2].New()
> look weird since I would expect:
>
> instance = itk.ClassITK(arg1, arg2)
> Would it make sense to always get an instance?

As of ITK 4.13, the behavior is

```
output = itk.ClassITK(arg1, arg2)
```

for ITK 5, maybe we should add lower case, snake functions for `itk::ProcessObject`'s for this  behavior:

```
output = itk.process_object_as_function(arg1, arg2, keyword_arg=value)
```

and just have

```
object = itk.ProcessObjectClass(arg1, arg2, keyword_arg=value)
```

just forward its arguments to `itk.ProcessObjectClass.New(*args, **kwargs)`. What do you think?

CC: @fbudin69500

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
```diff
@@ -27,95 +27,70 @@
...
+outputImageFile = sys.argv[3]
+differenceImageAfterFile = sys.argv[4]
+differenceImageBeforeFile = sys.argv[5]
+
 
 Dimension = 2
-PixelType = itk.F
+PixelType = itk.ctype('float')
 
 FixedImageType = itk.Image[PixelType, Dimension]
```
fbudin69500: `FixedImageType` and `MovingImageType` should be deduced from the loaded image if that is necessary, or these types could be removed. Same thing for `Dimension`: Dimension = fixedImage.GetImageDimension()

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
fbudin69500: See my comment above about `Dimension`, although deducing the dimension from the image is not a bad solution either.

fbudin69500:
Using camelcase for function calls is a great idea. I'll try to give it a shot to see how difficult it is to implement this.

src/Registration/Common/Perform2DTranslationRegistrationWithMeanSquares/Code.py
thewtex: In this case, since the filters involved are only available for floating point images, we have to specify the desired pixel type to have a floating point pixel.  Good point on the `fixedImage.GetImageDimension()` -- added.

PR #53, opened 2018-05-08
DOC: Migrate peer review description from Gerrit to GitHub

thewtex:


thewtex:
Thanks for the review, Jon!

Your thoughts on these PRs are appreciated.

I am also using the PRs to bootstrap testing of the new continuous integration, so I may merge them before they have had a chance for normal review.

jhlegarreta:
No worries. I understand. I think it's working great.

I've noticed the new checks; I guess it is still WIP, and that you're putting in place everything needed. I ignore why we are missing the circleCI or CDash builds in this repository.

Although, since a couple of days the PR's have a new tab called Checks where GitHub marketplace apps are supposedly able to provide information. I ignore how this can benefit our repos.

One thing that I think ITKExamples is missing, as opposed to the ITKSoftwareGuide repository, is to request reviews before merging for each PR, as mentioned in [this ISC card](https://github.com/orgs/InsightSoftwareConsortium/projects/1#card-9445023). I cannot do this due to my limited rights on the repostory.

thewtex:
Yes, CI-builds are on their way in #55 . After we get CI up, we could enable required reviews. It looks like it will be an incremental process -- all three platforms have build issues or test failures.  That's what happens when you don't have good regression testing :-P. We should be able to get Linux clean, and incrementally get Windows and macOS clean.

PR #56, opened 2018-05-11
ENH: adding a new example to demonstrate use of ParallelizeImageRegion

dzenanz:
To go along with multi-threading refactoring started in [this patch](http://review.source.kitware.com/#/c/23175/).


src/Core/Common/UseParallelizeImageRegion/Code.cxx
```diff
@@ -0,0 +1,116 @@
...
+          }
+      },
+      nullptr); //we don't have a filter whose progress needs to be updated
+}
+
+int main(int, char *[])
+{
+  int result = EXIT_SUCCESS;
+
+  //create an image
```
thewtex: please add a comment here that indicates what `{0, 0}` and `{50, 20}` are (the Index and Size)

src/Core/Common/UseParallelizeImageRegion/Code.cxx
```diff
@@ -0,0 +1,116 @@
...
+  duplicator->Update();
+
+  ImageType::Pointer clonedImage = duplicator->GetOutput();
+  clonedImage->DisconnectPipeline();
+
+  //invoke the two functions
+  log1x_LogImageFilter(image);
+  log1x_ParallelizeImageRegion(clonedImage);
+
+  //compare to make sure the results are the same
```
thewtex: use `Const` interators

src/Core/Common/UseParallelizeImageRegion/Documentation.rst
```diff
@@ -0,0 +1,32 @@
...
+--------
+
+
+This example demonstrates how to take advantage of MultiThreaderBase::ParallelizeImageRegion.
+A comparison is made with LogImageFilter for purpose of computing log(1+x), where x is pixel value.
+
+
+Results
+-------
+
```
thewtex: We could count the number of pixel that are different, then print to `stdout` *LogImageFilter and ParallelizeImageRegion generate the same result.*

src/Core/Common/UseParallelizeImageRegion/Documentation.rst
```diff
@@ -0,0 +1,32 @@
...
+.. index::
+   single: MultiThreaderBase
+
+Synopsis
+--------
+
+
+This example demonstrates how to take advantage of MultiThreaderBase::ParallelizeImageRegion.
+A comparison is made with LogImageFilter for purpose of computing log(1+x), where x is pixel value.
```
thewtex: We could provide more information here on why this is important, e.g.,

> With ``ParallelizeImageRegion``, we can process an image efficiently and in parallel with a function that is defined in a flexible way.  The entire operation, defined in 10 lines of code, previously required the definition of an entire class.

src/Core/Common/UseParallelizeImageRegion/Code.cxx
```diff
@@ -0,0 +1,116 @@
...
+  LogType::Pointer logF = LogType::New();
+  logF->SetInput(image);
+  logF->SetInPlace(true);
+  logF->Update();
+  image = logF->GetOutput();
+  image->DisconnectPipeline();
+}
+
+//calculate log(1+x), where x is pixel value, using ParallelizeImageRegion
+void log1x_ParallelizeImageRegion(ImageType::Pointer& image)
```
jhlegarreta: Is there any particular reason to use underscores in function names here? This is not consistent with the ITK style :worried:

src/Core/Common/UseParallelizeImageRegion/Code.cxx
```diff
@@ -0,0 +1,116 @@
...
+#include "itkRandomImageSource.h"
+#include "itkImageDuplicator.h"
+#include "itkImageRegionIterator.h"
+
+constexpr unsigned int Dimension = 2;
+using PixelType = unsigned int;
+using ImageType = itk::Image< PixelType, Dimension >;
+
+//calculate log(1+x), where x is pixel value, using LogImageFilter
+void log1x_LogImageFilter(ImageType::Pointer& image)
```
jhlegarreta: Is there any particular reason to use underscores in function names here? This is not consistent with the ITK style :worried:

src/Core/Common/UseParallelizeImageRegion/Code.cxx
```diff
@@ -0,0 +1,116 @@
...
+#include "itkLogImageFilter.h"
+#include "itkRandomImageSource.h"
+#include "itkImageDuplicator.h"
+#include "itkImageRegionIterator.h"
+
+constexpr unsigned int Dimension = 2;
+using PixelType = unsigned int;
+using ImageType = itk::Image< PixelType, Dimension >;
+
+//calculate log(1+x), where x is pixel value, using LogImageFilter
```
jhlegarreta: I know this is a matter of tastes, but the style guidelines state that we should leave a blank space between `//` and the comment sentence, and start with capitals.

src/Core/Common/UseParallelizeImageRegion/Code.cxx
```diff
@@ -0,0 +1,116 @@
...
+  image->DisconnectPipeline();
+}
+
+//calculate log(1+x), where x is pixel value, using ParallelizeImageRegion
+void log1x_ParallelizeImageRegion(ImageType::Pointer& image)
+{
+  itk::MultiThreaderBase::Pointer mt = itk::MultiThreaderBase::New();
+  //ParallelizeImageRegion invokes the provided lambda function in parallel
+  //each invocation will contain a piece of the overall region
+  mt->ParallelizeImageRegion<Dimension>(image->GetBufferedRegion(),
```
jhlegarreta: Again, this is more a matter of tastes, and may be a lost battle :sweat:, but the style guide states that we should leave a blank space between the template argument characters and the actual argument (`< Dimension >`). The same applies to function, control statement (`for( whatever )` instead of  `for (whatever)` etc.) arguments.

src/Core/Common/UseParallelizeImageRegion/Code.cxx
```diff
@@ -0,0 +1,116 @@
...
+  log1x_ParallelizeImageRegion(clonedImage);
+
+  //compare to make sure the results are the same
+  itk::ImageRegionIterator<ImageType> it1(image, region);
+  itk::ImageRegionIterator<ImageType> it2(clonedImage, region);
+  for (; !it1.IsAtEnd(); ++it1, ++it2)
+    {
+    if (it1.Get() != it2.Get())
+      {
+      std::cerr << "Pixel values are different at index "<<it1.GetIndex()
```
jhlegarreta: white spaces: `index " << it1.GetIndex()`

src/Core/Common/UseParallelizeImageRegion/Code.cxx
dzenanz: That fights against Visual Studio's default style. Manually correcting that is super-annoying. I do try to do that when changing existing code which uses that style.

thewtex: Is there a Visual Studio ITK style file?

dzenanz: In `for ( ;` there is a space before semicolon, which KWStyle does not like.

hjmjohnson: We should consider:  http://editorconfig.org

It provides a nice way to define styles that is supported by many editors.

Hans

src/Core/Common/UseParallelizeImageRegion/Documentation.rst
dzenanz: Done

src/Core/Common/UseParallelizeImageRegion/Documentation.rst
dzenanz: Done

src/Core/Common/UseParallelizeImageRegion/Code.cxx
dzenanz: ITK's style is so complicated there isn't a style tool which can enforce it in an automated and satisfactory way. Should we reconsider loosening it? Or changing it entirely, like VTK did? If so we should make it simple enough that we have some automated way of enforcing it. Should we start a Discourse discussion about this?

jhlegarreta:
👍  for Hans' proposal. MSVC seems to like `EditorConfig` files.

I ignore what process the VTK folks followed, but I do remember some hot discussions on the mailing list about it. I do not mind if we change it if we stick to a set of rules. One may find a style less readable than another, but if the toolkit is consistent, the reader ends up by getting used to it, and adopting it.

Due to some oversights, I remember having asked by VTK folks to modify the style of some of my contributions to conform to the rules of VTK. So there are surely aspects that cannot be enforced automatically. I do not remember VTK having a config file to do this across all compilers. But I remember MSVC asking me about this. Although it may seem that the effort is not worthwhile, we remember about it when reading the code.

I've already assumed that the white space between the brackets and the method arguments is a lost battle.

No arguments against discussing it on discourse.

blowekamp:
Your selected example can be trivially implemented with the new UnaryGeneratorImageFilter. It would really make an excellent illustration of using a per-pixel lambda function. It would not be good to mislead, that this is the best way to implement this type of per-pixel operation.

I suggest something slightly more complicated for this example, perhaps using a neighborhood.

dzenanz:
To make a point, the example should be short. Involving a neighborhood will make it noticeably longer. Do you have a concrete suggestion?

thewtex:
The example should not be more complicated, which distracts from the purpose of the example.

Manual use of UnaryGeneratorImageFilter could also be shown, but the ParallelizeImageRegion example should be retained. There are many cases where just calling ParallelizeImageRegion is preferred because of the added flexibility and reduced overhead relate to a filter class.

blowekamp:
Something **slightly** more complicated than what can be trivially implemented with the per-pixel Generator/Functor filters could be something that accesses the index/point.

```

```

dzenanz:
Well, I will just create another example which only uses `ParallelizeImageRegion`, and does something simple but non-trivial. Should we keep this example in addition to that new one?

blowekamp:
The new threading methods are certainly important enough to warrant multiple examples.

dzenanz:
I am not sure I got the external data correctly, can someone double check that?

thewtex:
> I am not sure I got the external data correctly, can someone double check that?

It looks good. After updating ITK, the CI builds will verify.

thewtex:
Verified locally.

PR #58, opened 2018-05-25
ENH: Bump Superbuild ITK to v5.0a02

thewtex:


thewtex:
@dzenanz @blowekamp A few updates to the resample test baselines were required:

  https://open.cdash.org/testDetails.php?test=657027960&build=5392967

  https://open.cdash.org/testDetails.php?test=657027964&build=5392967

thewtex:
This is with linear interpolation.

I don't think a floating point precision flag is feasible or desirable.

PR #68, opened 2018-09-16
ENH: Add Azure Pipelines configuration

thewtex:


jhlegarreta:
Awesome, I can see the azure build progress and live output.

jhlegarreta:
- *0 Compiler errors*
- *50 or more Compiler warnings*

The job could not be submitted because the drop site seems to be empty.

thewtex:
Nice @jhlegarreta ! Thanks for the update.

I am going to get an initial configuration going, then merge, then try to re-enable using an `itkrobot` account...

PR #69, opened 2018-09-19
ENH: Move Azure DevOps Pipeline YAML file into Utilites/

thewtex:
Avoid top level clutter.


thewtex:
@jcfr another excellent use of the `ninja` Python package! :clap:

PR #71, opened 2018-09-19
DOC: Add Azure Pipelines CI Status badge

thewtex:


thewtex:
> HistogramCreationAndBinAccessTestPython

@dzenanz has fixed some bugs recently in this arena. @dzenanz any ideas?

dzenanz:
I was modifying Otsu filter, and this is plain and straight histogram manipulation. Also, C++ and Python code seem equivalent so it is strange only Python test fails.

thewtex:
Rebasing on master...

PR #81, opened 2018-11-13
ENH: Update the CMake minimum required version.

jhlegarreta:
Update the CMake minimum required version to match the one required by
ITKExamples master.


jhlegarreta:
There are a quite few build failures complaining about HDF5, and many other test failures that will need to be addressed in another PR.

Merging since they are unrelated to this topic.

PR #82, opened 2018-11-16
COMP: Replace vnl_math_abs with std::abs

thewtex:
Support for vnl_math_abs has been removed in ITK 5.


thewtex:
CI error addressed in #83

hjmjohnson:
@thewtex Thanks.

PR #83, opened 2018-11-17
COMP: Run apt-get update in Linux Azure build

thewtex:
To address missing packages in python3-venv install.


thewtex:
@jhlegarreta thanks for the review :+1:

Issue #23, opened 2014-07-22
Superbuild does not pass Sphinx output options.


thewtex:
Does not pass

SPHINX_HTML_OUTPUT:BOOL=ON
SPHINX_LATEX_OUTPUT:BOOL=ON
SPHINX_EPUB_OUTPUT:BOOL=ON

-> we are not getting these outputs on the Download page.

PR #84, opened 2018-12-08
COMP: fix CMake configure warning when Python is not found

dzenanz:
If Python is not found, PYTHON_VERSION_STRING is empty causing:

CMake Error at src/CMakeLists.txt:26 (if):
  if given arguments:

    "VERSION_LESS" "2.7"

  Unknown arguments specified


src/CMakeLists.txt
```diff
@@ -19,7 +19,9 @@ endif()
 
 if(BUILD_DOCUMENTATION)
   find_package(PythonInterp REQUIRED)
-
+  if(NOT PYTHONINTERP_FOUND)
+    message(SEND_ERROR "Building the documentation requires Python")
```
hjmjohnson: @dzenanz A send error will continue processing.  I think line 28 needs amending as well.

perhaps
line #23:  add "no python interpreter found" to message.
line 28:  if( NOT PYTHONINTERP_FOUND  OR ...)

dzenanz: I didn't realize that `SEND_ERROR` will continue processing. I will amend the commit soon.

PR #91, opened 2019-03-14
Fix sphinxexamples build 20190314

hjmjohnson:
Attempt to build Sphinx examples as an internal project as part of ITK proper.  This is the first step needed to include these examples as an updated set of reference examples that demonstrate best practices uses of ITK.


hjmjohnson:
FYI:  The ITK remote to ITKExamples does not currently build with the very old matched release version.   This needs to be updated in ITK/Module/Remote.

hjmjohnson:
Failures seem to be unrelated to this PR.  I think the failures are due to broken testing.

PR #111, opened 2019-06-10
Fix itk internal build

hjmjohnson:
Compilation changes needed in cases of internal ITK build & SuperBuild scenarios.

Resolved target conflicts when building ITK/Examples and SphinxExamples.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -8,7 +8,9 @@ find_package(ITK REQUIRED)
 include(${ITK_USE_FILE})
 
 # To avoid linker issues.
-if(ITKVtkGlue_LOADED)
+# Module_ITKVtkGlue is true when building inside ITK
+# ITKVtkGlue_LOADED is true when SuperBuild ITKEx stand-alone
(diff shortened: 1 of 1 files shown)


src/Core/ImageFunction/CMakeLists.txt
```diff
@@ -4,8 +4,9 @@ compare_to_baseline(EXAMPLE_NAME MultiplyKernelWithAnImageAtLocation
   BASELINE_PREFIX OutputBaseline
   )
 
-add_example(GaussianBlurImageFunction)
-compare_to_baseline(EXAMPLE_NAME GaussianBlurImageFunction
+# TODO: Evaluate if GaussianBlurImageFunctionEx duplicates the Examples/Filtering/CMakeLists.txt:GaussianBlurImageFunction
```
dzenanz: https://github.com/InsightSoftwareConsortium/ITK/blob/master/Examples/Filtering/GaussianBlurImageFunction.cxx and https://github.com/InsightSoftwareConsortium/ITKExamples/blob/master/src/Core/ImageFunction/GaussianBlurImageFunction/Code.cxx seem to be duplicates. We should probably remove one of them.

PR #114, opened 2019-06-24
ENH: Add AdditiveGaussianNoiseImageFilter Example

wilson-ben:
Adding an example utilizing the AdditiveGaussianNoiseImageFilter.


src/Filtering/ImageFeature/AdditiveGaussianNoiseImageFilter/Code.cxx
```diff
@@ -0,0 +1,78 @@
...
+  if( argc != 5 )
+    {
+    std::cerr << "Usage: "<< std::endl;
+    std::cerr << argv[0];
+    std::cerr << " <InputFileName> <OutputFileName> [Mean] [Standard Deviation]";
+    std::cerr << std::endl;
+    return EXIT_FAILURE;
+    }
+    // Initialize and assign user provided variables
+  const char * inputImage = argv[1];
```
hjmjohnson: Can these be std::string?

src/Filtering/ImageFeature/AdditiveGaussianNoiseImageFilter/Code.cxx
```diff
@@ -0,0 +1,78 @@
...
+  const char * outputImage = argv[2];
+  // get floating point numbers for the Mean and Standard Deviation to perform the algorithm
+  const double mean = atof(argv[3]);
```
hjmjohnson: Avoid atof. 
Prefer error checked std::stod over atof

wilson-ben: Will change, thanks

src/Filtering/ImageFeature/AdditiveGaussianNoiseImageFilter/Code.cxx
wilson-ben: Yes, I will change them to std::string

jhlegarreta: @hjmjohnson @thewtex sorry to chime in but I'm wondering whether the same criterion is used in ITK. If not, and if it should be used, we may need to open an issue. Thanks.

src/Filtering/ImageFeature/AdditiveGaussianNoiseImageFilter/Code.cxx
```diff
@@ -0,0 +1,78 @@
...
+  if( argc != 5 )
+    {
+    std::cerr << "Usage: "<< std::endl;
+    std::cerr << argv[0];
+    std::cerr << " <InputFileName> <OutputFileName> [Mean] [Standard Deviation]";
+    std::cerr << std::endl;
+    return EXIT_FAILURE;
+    }
+    // Initialize and assign user provided variables
+  const std::string * inputImage = argv[1];
```
hjmjohnson: @wilson-ben This code is now wrong.

`const std::string *`
`const std::string`

jhlegarreta: Change to `const char *` in 5c9b8e7.

jhlegarreta:
@wilson-ben reviews are appreciated.

PR #115, opened 2019-06-28
STYLE: Prefer error checked std::sto[id] over ato[if]

jhlegarreta:
The ato[if] functions do not provide mechanisms for distinguishing
between '0' and the error condion where the input can not be converted.

std::sto[id] provides exception handling and detects when an invalid
string attempts to be converted to an [integer|double].

ato[if]()
   Con: No error handling.
   Con: Handle neither hexadecimal nor octal.

The use of ato[if] in code can cause it to be subtly broken.
ato[if] makes two very big assumptions indeed:
   The string represents an integer/floating point value.
   The integer can fit into an int.

### DIFF
--- a/src/Compatibility/Deprecated/ResampleRGBImage/Code.cxx
+++ b/src/Compatibility/Deprecated/ResampleRGBImage/Code.cxx
@@ -50,7 +50,7 @@ int main(int argc, char *argv[])
         input = reader->GetOutput();
         if (argc > 2)
         {
-            factor = atof(argv[2]);
+            factor = std::stod(argv[2]);
         }
     }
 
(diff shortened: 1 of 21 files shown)


jhlegarreta:
CI failure is unrelated:
```
CMake Error at src/CMakeLists.txt:20 (message):
  ITKVtkGlue required for SphinxExamples
```
as we're experiencing it lately in other PRs.

I'd dare to say that merging this is safe.

jhlegarreta:
CI failure is unrelated:
```
CMake Error at src/CMakeLists.txt:20 (message):
  ITKVtkGlue required for SphinxExamples
```

as we're experiencing it lately in other PRs.

I'd dare to say that merging this is safe.

Issue #22, opened 2014-05-23
CreateNewExample should automatically use the ITK source tree used for the build


thewtex:

jhlegarreta:
Definitely #101 did not address this. We need to see if we manage to address it in #120.

PR #120, opened 2019-06-29
WIP: BUG: Fix new example script cookiecutter

jhlegarreta:
Fix new example script cookiecutter: launch the cookiecutter using a
Python script:
- Remove the `cookiecutter.json`: it is generated dynamically by the
  script.
- Use a project post-generation cookiecutter hook to add the necessary
  information to the ITK group and modules the class being demonstrated
  belongs to, and to move the project generated by cookiecutter to the
  appropriate place in the ITKExamples source tree.


jhlegarreta:
@thewtex I think we can use Python itself to make the pre-commit hook modify the `cookiecutter.json` file with the `group_name` and `module_name` values that the hook detected: we could modify them using `pysed, `sh.sed` or `re.sub`, for example, and thus avoid to try to inject content into the cookiercutter (and hence, avoid the infinite loop). May be not perfect/seem hacky, but could be a valid strategy unless we find a better one. Have not tested it yet.

Any thoughts?

jhlegarreta:
I did a few tests, and it looks like things are not easy the way they are posed right now.

In summary, it looks like the cookiecutter reads/loads the json file once, before the python script is executed, thus any attempt to modify the JSON file is useless.

So, I'd dare to say that we have just two options:

1. **Do not use the pre-gen hook, and let the `group_name` and `module_name` cookiecutter template variables have the values provided by the user**, as suggested by @thewtex, **and use a post-gen hook just** (i.e. no `get_group_and_module_from_class_name` function) **to move the example in the tree structure** according to the values provided by the user.
  **Pros**:
  - We are calling the cookiecutter directly as for the ITK remote modules, which may sound fmailiar.
  - The user is prompted for every value, which is intuitive.
  **Cons**:
  - If users to write the `class_name`, `group_name`, and `module_name` without typos or errors, theY will need to first locate where the example was created, if the cookiecutter managed to do so, and then fix the typos/errors manually.
  - It seems counter-intuitive to me to first create some structure, and then move it.

2. **Use a python script to launch the cookiecutter**, as suggested [here](https://github.com/InsightSoftwareConsortium/ITKExamples/pull/101#issuecomment-497900225). **The python script would take as its arguments the variables required by the cookiecutter**, and would either create a JSON file for the cookiecutter context or just provide them in the script. Creating a JSON file may seem an unnecessary step/counter-intuitive, as it will stay there and be detected by git as a new file.
  **Pros**:
  - We are guaranteed to create the example at the appropriate place (the script will check if the class exists in the ITK source tree and prompt the user until it succeeds).

We can even make the script prompt the user, much like cookicutter does, so that the mechanism resembles a direct use of cookiecutter.

@thewtex Do you have other suggestions?

jhlegarreta:
13a4ad4 implements option 2.

**Tested** locally with Python 2.7.12. Working :tada: :fireworks: .

A minor comment:
The module and examples lists in
```
group/CMakeLists.txt
group/index.rst
```
and
```
group/module/CMakeLists.txt
group/module/index.rst
```

will not be alphabetically ordered since we are appending contents. In fact, they are currently not alphabetically ordered.

But can be done programmatically if needed.

jhlegarreta:
4ee7c47 fixes a bug that went unnoticed in the previous commit. Sorry.

jhlegarreta:
Next step may be to avoid asking for the ITK and ITKExamples sources. The latter should be easy. The former has to do with #22. Suggestions are welcome.

thewtex:
> [13a4ad4](https://github.com/InsightSoftwareConsortium/ITKExamples/commit/13a4ad494583b5cf5e4249c615623547fc16e607) implements option 2.
> 
> **Tested** locally with Python 2.7.12 and 3.6.5. Working.

Awesome!!! :taco: :tada: :fireworks: 



>     1. **Use a python script to launch the cookiecutter**, as suggested [here](https://github.com/InsightSoftwareConsortium/ITKExamples/pull/101#issuecomment-497900225). **The python script would take as its arguments the variables required by the cookiecutter**, and would either create a JSON file for the cookiecutter context or just provide them in the script.

This is a great approach.



> will not be alphabetically ordered since we are appending contents. In fact, they are currently not alphabetically ordered.

I think it is fine that they are not alphabetically sorted. In fact, it would be better if they are sorted by topic so a reader can page forward and backward through examples and see similar examples.  Appending to the end is a good start. We can sort as an improvement, if it is helpful, after the example is complete.



> Next step may be to avoid asking for the ITK and ITKExamples sources. The latter should be easy. The former has to do with #22. Suggestions are welcome.

Nice!

The source directory could be found by inspecting `ITK_SOURCE_DIR` in the build directory *CMakeCache.txt*.


Amazing work @jhlegarreta -- thanks!

jhlegarreta:
Thanks for the review @thewtex  |

> The source directory could be found by inspecting ITK_SOURCE_DIR in the build directory CMakeCache.txt.

Am I wrong or the build directory is yet unknown to the script ?

thewtex:
> Am I wrong or the build directory is yet unknown to the script ?

Yes -- may be do not require a build directory? Or, maybe we put the script in and examples build directory, and find the default locations of the ITK source and build trees assuming their superbuild locations?

jhlegarreta:
> may be do not require a build directory? Or, maybe we put the script in and examples build directory, and find the default locations of the ITK source and build trees assuming their superbuild locations?

Looks like that is the only option we are left with to locate the ITK source automatically.

jhlegarreta:
@thewtex I've been investigating this, and I'd dare to say that further thinking/changes would be needed.

The thing is that the `Utilities` folder [gets copied](https://github.com/InsightSoftwareConsortium/ITKExamples/blob/4a267bb80d2e11e4b6ee47cb023eac47e8e728f9/CMakeLists.txt#L190) if the flag `BUILD_DOCUMENTATION` is set to `ON`.

In the case of the script we are trying to put in place, if I'm not wrong we would require a call to `configure_file` in the `CMakeLists.txt`, regardless of the `BUILD_DOCUMENTATION` flag.

Would that mean that we should prevent the `CreateNewExample.py.in` file from being copied? How would we do that without manually listing all the files in the `Utilities` folder?

thewtex:
Can we change it so that *CreateNewExample.py* is always generated from *CreateNewExample.py.in* regardless of the `BUILD_DOCUMENTATION` setting? We could move the Python script in the repository tree if needed.

jhlegarreta:
I was testing locally having added the `*.in` suffix, catching automatically the ITK and ITKEx source paths. 

As for changing the location of the file, I think `Utilities` is the right one. May be I can list the files in the directory, excluding it with the `list(FILTER "${ITKExamples_SOURCE_DIR}/Utilities" EXCLUDE "*.in")` CMake command, then looping over the list to copy the files, and adding the `configure_file` for the `CreateNewExample.py.in` right after the `BUILD_DOCUMENTATION` `if` conditional block?

jhlegarreta:
A few comments about e4a1337:
- I am not seeing that the `Utilites` folder is being copied in the tests I have been doing for the last few hours, so I have not filtered the folder contents by now. **To be checked/confirmed**.
- I am **not getting the `CookieCutter` folder to be copied to the build directory**, although I have supposedly added the [corresponding commands](https://github.com/InsightSoftwareConsortium/ITKExamples/commit/e4a1337458fde8398066fc79b3ef57048fd0969c#diff-af3b638bc2a3e6c650974192a53c7291R319). Fixes are **welcome**.
- Manually copied the folder to verify that the scripts directories are OK and that it is working as expected. It turns out that the **cookiecutter post gen hook complains** [here](https://github.com/InsightSoftwareConsortium/ITKExamples/commit/e4a1337458fde8398066fc79b3ef57048fd0969c#diff-68ead466ce19f3c9b310b1d3d0e68811R216). No idea why. Tested on Windows 10, using a conda virtualenv and Python 3.6.0 this time.

jhlegarreta:
BTW, I did not use `ITK_SOURCE_DIR`, because that would mean:
1. Looking for `ITK_DIR` in the ITKExamples build `CMakeCache.txt`.
2. Then looking for `ITK_SOURCE_DIR` in the `ITK_DIR` `CMakeCache.txt`.
3. And even if that indeed gives me the ITK source, following #22, we need a guarantee that that source comes from the Superbuild.

The only rationale I found to guarantee that was to [just check](https://github.com/InsightSoftwareConsortium/ITKExamples/pull/120/commits/e4a1337458fde8398066fc79b3ef57048fd0969c#diff-5f126807a2835c0ae26937712039c5ddR169) if an `ITK` directory exists in the ITKExamples build directory.

Comments are welcome.

thewtex:
Can we use `ITK_SOURCE_DIR` from the superbuild, if found, and prompt the user for an ITK source directory location, otherwise?

jhlegarreta:
@thewtex can you provide some more insight to do so?

I do not see the way to get the superbuild. In my understanding, we will always get some value for `ITK_SOURCE_DIR`, and I do not see a way to tell if it comes from the superbuild or not, other than the very weak assumption I made.

Also, can you please have a look at why the `CookieCutter` dir copy commands are not doing their job?

Thanks !

thewtex:
@jhlegarreta yes, I will dive in, but I am not going to get a chance until later next week.

jhlegarreta:
@thewtex I've run out of ideas, so that would be very welcome. We're getting closer.

jhlegarreta:
@thewtex Should we re-consider this/give it another try?

thewtex:
@jhlegarreta yes -- could this branch please be re-based on `master`?

jhlegarreta:
681e9cc rebased on `master`.

thewtex:
@jhlegarreta thanks!

thewtex:
Outstanding work on this topic @jhlegarreta ! Amazing job!

> I do not see the way to get the superbuild. In my understanding, we will always get some value for `ITK_SOURCE_DIR`, and I do not see a way to tell if it comes from the superbuild or not, other than the very weak assumption I made.

The tooling correctly gets the source directory from the superbuild, but we do not need to guarantee that it *must* be from the superbuild.



> Also, can you please have a look at why the `CookieCutter` dir copy commands are not doing their job?

I made a few fixes:

- In *CreateNewExample.py*, use the *CookieCutter* directory as located relative to the *CreateNewExample.py* script
- When copying the *CookieCutter* sources, use *add_custom_target* with the *ALL* option, and add the missing *Utilities* directory.

Merging so @hina-shah and others can continue their good work!

jhlegarreta:
Have not tested but thanks @thewtex !

@hjmjohnson @mseng10 @hina-shah sorry that this took so long. Checking that the new generation system works as expected would be greatly appreciated.

PR #149, opened 2019-12-02
ENH: Add SetDefaultNumberOfThreads

thewtex:
Addresses: https://github.com/InsightSoftwareConsortium/ITK/issues/113

### DIFF
--- a/src/Core/Common/CMakeLists.txt
+++ b/src/Core/Common/CMakeLists.txt
@@ -204,3 +204,5 @@ compare_to_baseline(EXAMPLE_NAME OutOfBoundsPixelsReturnConstValue
   BASELINE_PREFIX OutputBaseline
   )
 endif()
+
+add_example(SetDefaultNumberOfThreads)
--- a/src/Core/Common/SetDefaultNumberOfThreads/CMakeLists.txt
+++ b/src/Core/Common/SetDefaultNumberOfThreads/CMakeLists.txt
@@ -1,0 +1,36 @@
+cmake_minimum_required(VERSION 3.10.2)
+
+project(SetDefaultNumberOfThreads)
+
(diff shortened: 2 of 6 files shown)


thewtex:
@mrocklin please take a look

mrocklin:
From a Python user's perspective I would prefer an API like ...


```python
import itk
itk.set_nthreads(4)
```

There are a few other libraries (blosc comes to mind) that do something similar.

However, with proper documentation what's here also works. I imagine that the sorts of people who are going to want to change the default number of threads are also going to have a decently high tolerance for writing lots of code.

Thank you all for working on this.

mrocklin:
Yeah, that was my guess as to what was happening.  Thought I'd mention it anyway.  Thanks all!

dzenanz:
It should not be hard to implement. A stand-alone function `itk::set_nthreads`:
```C++
void 
set_nthreads(ThreadIdType numberOfThreads)
{
  itk::MultiThreaderBase::SetGlobalDefaultNumberOfThreads(numberOfThreads);
}
```
should be about enough.

dzenanz:
While such a function wouldn't make much sense in C++, we could make it available only in Python. Thoughts?

thewtex:
Thanks for the reviews!

The CI failures are unrelated, so merging.

If there is demand for it, we could add `itk.set_nthreads` by defining it in [`itkExtras.py`](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Wrapping/Generators/Python/itkExtras.py).

It would look like:

```
def set_nthreads(number_of_threads):
    threader = itk.MultiThreaderBase.New()
    threader.SetGlobalDefaultNumberOfThreads(number_of_threads)
```

Issue #139, opened 2019-09-30
SegmentWithWatershedImageFilterTestPython1 failing test


mseng10:
This issue is derived from #131 to keep track of indivial examples failing tests.

The following tests fail with ITK's latest release: 

- SegmentWithWatershedImageFilterTest01Python

- SegmentWithWatershedImageFilterTest01BaselineComparisonPython

The issue arrises on line 76 of src/Segmentation/Watersheds/SegmentWithWatershedImageFilter when the colormapImageFilter object attempts initialize its enumeration value. Due to the recent transition from 'c' style enums to enum class objects in source ITK, many enums were changed and moved. 

The enumeration: `RGBColormapFilterEnumType`  was created and moved outside its class: `ScalarToRGBColormapImageFilter` in source ITK. This enumeration was not wrapped and thus cannot be accessed in ITKPythonPackage 5.1b1. More information can be viewed in HEAD issue #131. 

When these enumerations are properly wrapped and implemented into the ITKPythonPackage, the failing test should resolve. Subsequently, the BaselineComparison test should pass as the error should not affect the resulting image.

mseng10:
Fixed in #156. Closing now.

Issue #131, opened 2019-09-20
Incomplete python wrapping of enum class objects


mseng10:
### Description

In the most recent set of failing ITKExamples tests, #130, the listed tests are unsolvable unless changed from the source. In the latest release of the ITKPythonPackage 5.1b.1 errors occur when attempting to access python wrapped enumerations. 

The issue arises because some enumerations are no longer wrapped after the 
#1061 pull request (In ITK). In #1061, they were changed from 'c' style enumerations to enum classes. During this transition a number of new enums were moved outside their original classes.

### Steps to Reproduce

1. Clone the latest version of ITKExamples from Github: 

`git clone -b master https://github.com/InsightSoftwareConsortium/ITKExamples.git`

3. Make build directory:  `mkdir ITKExamples-Superbuild`

4. Allocate to build directory:  `cd ITKExamples-Superbuild`

5. Run:  `cmake ../ITKExamples/Superbuild`

6. Configure CMake with:  `ITKExamples_USE_WRAP_PYTHON = ON`

7. Run: `make`

8. Allocate to ITKEx-build directory: `cd ITKEx-build`

9. Test: `ctest -V`
 

### Expected behavior

100% Tests passed.

### Actual behavior

```   
98% tests passed, 12 tests failed out of 589
248 - ApplyAColormapToAnImageTestBaselineComparisonPython (Failed)
252 - ApplyAColormapToAnImageTestPython (Failed)
568 - SegmentWithWatershedImageFilterTest01BaselineComparisonPython (Failed)
570 - SegmentWithWatershedImageFilterTest02BaselineComparisonPython (Failed)
572 - SegmentWithWatershedImageFilterTest03BaselineComparisonPython (Failed)
574 - SegmentWithWatershedImageFilterTest04BaselineComparisonPython (Failed)
576 - SegmentWithWatershedImageFilterTest05BaselineComparisonPython (Failed)
583 - SegmentWithWatershedImageFilterTest01Python (Failed)
584 - SegmentWithWatershedImageFilterTest02Python (Failed)
585 - SegmentWithWatershedImageFilterTest03Python (Failed)
586 - SegmentWithWatershedImageFilterTest04Python (Failed)
587 - SegmentWithWatershedImageFilterTest05Python (Failed)
```

### Reproducibility
The tests fail 100% of the time when the ITKPythonPackage 5.1b.1 is used. 

When ITKPythonPackage 5.0.1 is used, 100% of tests pass.

[...]

PR #156, opened 2020-04-30
BUG: Fix Failing module tests

mseng10:
This PR fixes all failing tests for ITKExamples.

Many of the python scripts needed to be refactored to incorporate the new ITKv5 structure. This required updating enumerations and changing `SetNumberOfThreads` to `SetNumberOfWorkUnits`.

`DelaunayConformEdgeFlipping` needed a new input image as the original wave.vtk did not process correctly. To make sure it was due to the image, I tried testing it in source ITK's DelaunayConformEdgeFlippingTest, and it produced the same error. I replaced it will mushroom.vtk which is used in source ITK's DelaunayConformEdgeFlippingTest.

`ReadAndPrintDICOMTags` needed an exception added for Specialized Values that are unable to be processed at runtime.

`RegisterIOFactories` needed its return statement deleted to allow the rest of the test to finish. The exception called is to be expected.

`AdditiveGaussianNoiseImageFilter` needed its I/O data transferred to Girder. In addition, a python test for its `Code.py` was added. I removed its `compare_to_baseline` test, because no OutputBaseline was present in the directory. When I tried configuring my own, I was unable to get it to pass despite the example working as intended.


mseng10:
This addresses issues: #131, #138,  #139, #140, #141, #142, #143.

hjmjohnson:
@mseng10 Nice work!

PR #161, opened 2020-06-01
Build documentation with recent Sphinx

thewtex:


dzenanz:
Is the documentation also supposed to build with older versions of Sphinx?

dzenanz:
Also, which version of Sphinx does it build with now? I missed that if it was mentioned somewhere in the PR.

PR #168, opened 2020-07-18
ENH: Add Python code for `GetImageSize` example

jhlegarreta:
Add Python code for `GetImageSize` example.

### DIFF
--- a/src/Core/Common/GetImageSize/Code.py
+++ b/src/Core/Common/GetImageSize/Code.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+
+# Copyright NumFOCUS
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import itk
+
+if len(sys.argv) != 2:
+    print('Usage: ' + sys.argv[0] + ' <inputImageFile>')
+    sys.exit(1)
+
+inputImageFile = sys.argv[1]
+
+image = itk.imread(inputImageFile, itk.UC)
+
+region = image.GetLargestPossibleRegion()
+size = region.GetSize()
+
+print(size)
+
+# Equivalently
+size = itk.size(image)
+
+print(size)
+
+# Corresponds to the NumPy ndarray shape. Note that the ordering is reversed.
+print(image.shape)
+
+# An example image had w = 200 and h = 100
+# (it is wider than it is tall). The above output
+# 200 x 100
+# so w = GetSize()[0]
+# and h = GetSize()[1]
+
+# A pixel inside the region
+indexInside = itk.Index[2]()
+indexInside[0] = 150
+indexInside[1] = 50
+print(region.IsInside(indexInside))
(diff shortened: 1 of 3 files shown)


jhlegarreta:
A few notes:
- Have not tested locally.
- Have preferred to follow the existing Python examples' conventions wrt to variable naming (i.e. camel case vs. snake case). Maybe at some point we may consider changing this to use a more Pythonic style and enforce it using a PEP checker.
- Have preferred to follow the existing Python examples' conventions wrt to the Python interface beyond the image reading (i.e. have not used the [procedural interface](https://discourse.itk.org/t/itk-5-0-beta-1-pythonic-interface/1271)/I ignore whether e.g. `get_largest_possible_region` exists). Although I believe that is an effort that might be suitable for a separate, massive PR, suggestions as appropriate are welcome.
- Although the comments in the `CXX` version seem a little awkward to me, I've kept them in the `Python` version.

Last but not least:
- The copyright lines supression range for Python files in the `Documentation.rst` file seems to be removing the `import sys` when it exists (and maybe the `import itk` when the former is not prsent). Not sure if this is intended or else this should be investigated and possibly changed for every Python file.

jhlegarreta:
Pinging @mseng10 .

src/Core/Common/GetImageSize/Documentation.rst
```diff
@@ -33,6 +33,13 @@ C++
 .. literalinclude:: Code.cxx
    :lines: 18-
 
+Python
+......
+
+.. literalinclude:: Code.py
+   :language: python
+   :lines: 1, 18-
```
dzenanz, line 41: This could be a copy-paste mistake. @thewtex?

thewtex: Since `17-` is where the code starts, it should be `17-` instead of `18-`. The `1,` keeps the script she-bang.

src/Core/Common/GetImageSize/Code.py
```diff
@@ -0,0 +1,57 @@
...
+# limitations under the License.
+
+import sys
+import itk
+
+if len(sys.argv) != 2:
+    print('Usage: ' + sys.argv[0] + ' <inputImageFile>')
+    sys.exit(1)
+
+Dimension = 2
```
thewtex, line 24: We can remove this for brevity.

src/Core/Common/GetImageSize/Code.py
```diff
@@ -0,0 +1,57 @@
...
+
+PixelType = itk.UC
```
thewtex, line 26: We can remove this for brevity.

src/Core/Common/GetImageSize/Code.py
```diff
@@ -0,0 +1,57 @@
...
+
+inputImageFile = sys.argv[1]
+
+image = itk.imread(inputImageFile, PixelType)
+
```
thewtex, line 31: While we can keep these below, we can also add,

```
# Equivalent
size = itk.size(image)

print(size)

# Corresponds to the NumPy ndarray shape. Note that the ordering is reversed.
print(image.shape)

src/Core/Common/GetImageSize/Code.py
jhlegarreta: Done in f0247e0.

src/Core/Common/GetImageSize/Code.py
jhlegarreta: Done in f0247e0.

src/Core/Common/GetImageSize/Code.py
jhlegarreta: Nice ! Done in f0247e0.

src/Core/Common/GetImageSize/Documentation.rst
jhlegarreta: Done in f0247e0.

PR #169, opened 2020-07-18
ENH: Add Python code for `GetTypeBasicInformation` example

jhlegarreta:
Add Python code for `GetTypeBasicInformation` example.

### DIFF
--- a/src/Core/Common/GetTypeBasicInformation/CMakeLists.txt
+++ b/src/Core/Common/GetTypeBasicInformation/CMakeLists.txt
@@ -21,3 +21,9 @@ install( FILES Code.cxx CMakeLists.txt
 enable_testing()
 add_test( NAME GetTypeBasicInformationTest
   COMMAND ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/GetTypeBasicInformation )
+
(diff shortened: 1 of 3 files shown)


jhlegarreta:
Same comments as for PR #168:
- Have not tested locally.
- Have preferred to follow the existing Python examples' conventions wrt to variable naming (i.e. camel case vs. snake case). Maybe at some point we may consider changing this to use a more Pythonic style and enforce it using a PEP checker.
- Have preferred to follow the existing Python examples' conventions wrt to the Python interface beyond the image reading (i.e. have not used the [procedural interface](https://discourse.itk.org/t/itk-5-0-beta-1-pythonic-interface/1271)/I ignore whether e.g. `is_negative` exists). Although I believe that is an effort that might be suitable for a separate, massive PR, suggestions as appropriate are welcome.

Last but not least:
- The copyright lines supression range for Python files in the `Documentation.rst` file seems to be removing the `import sys` when it exists (and maybe the `import itk` when the former is not prsent). Not sure if this is intended or else this should be investigated and possibly changed for every Python file.

jhlegarreta:
Pinging @mseng10

PR #173, opened 2020-07-22
ENH: Point `build, test` CI badge to builds

jhlegarreta:
Point `build, test` CI badge to builds: point to the `master` branch
status by default.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,8 @@
 ITK Examples
 ============
 
-.. image:: https://github.com/InsightSoftwareConsortium/ITKExamples/workflows/Build,%20test/badge.svg
+.. image:: https://github.com/InsightSoftwareConsortium/ITKExamples/workflows/Build,%20test/badge.svg?branch=master
+   :target: https://github.com/InsightSoftwareConsortium/ITKExamples/actions?query=workflow%3A%22Build%2C+test%22+branch%3Amaster
 
 .. image:: https://img.shields.io/badge/License-Apache%202.0-blue.svg?style=shield
    :target: https://github.com/InsightSoftwareConsortium/ITKExamples/blob/master/LICENSE


jhlegarreta:
@mseng10 @hjmjohnson thanks for the monumental effort to make the GitHub actions work :100: .

thewtex:
Was the correct CI configuration merged? There should not be Python / packaging builds.

mseng10:
@thewtex I configured 2 separate jobs. The first tests only `cxx`, while the second enables both `cxx` and Python. If we wanted to remove the `cxx` builds, I suppose the same range of testing would still be accomplished. 

The Python jobs have no actual Python wrapping, while instead tells the ITKExamples build that `ITK_WRAP_PYTHON` is enabled. This allows all the python tests to run, and be tested against the installed python package: `5.1.0.post2`.

jhlegarreta:
@thewtex good catch. Maybe the workflow could be renamed to just `Build, test` to avoid confusion?

thewtex:
@mseng10 nice! @jhlegarreta 's suggestion would help in this case.

jhlegarreta:
@mseng10 Do you mind if I let you change the workflow name and associated parts (probably just the badge in the `README`) and once that gets merged if I rebase this PR myself or do you prefer me to take care of the change?

jhlegarreta:
> @mseng10 Do you mind if I let you change the workflow name and associated parts (probably just the badge in the README) and once that gets merged if I rebase this PR myself or do you prefer me to take care of the change?

@mseng10 nevermind :pray:, I went ahead and renamed the workflow in #177. Reviews are welcome. Thanks again for the hard work with the Examples !

jhlegarreta:
Not sure why the badges would appear on separate lines. If no one has any insight on this, and unless the patch set breaks something, this is ready to be merged.

jhlegarreta:
Checks passing ! Merging.

PR #174, opened 2020-07-22
BUG: Fix failing tests reported by CI

jhlegarreta:
Fix failing tests reported by CI, e.g. in this buid:
https://github.com/InsightSoftwareConsortium/ITKExamples/runs/895512782?check_suite_focus=true

Fixes:
- Convert the returned values of the calls to string so that they can be
  printed properly in the kernel creation examples.
- Remove the brackets to get the `NumericTraits` attribute values in the
  examples calling its attributes.

Take advantage of the commit to add the missing Python test for the
`CreateForwardDifferenceKernel` example.

### DIFF
--- a/src/Core/Common/GetTypeBasicInformation/Code.py
+++ b/src/Core/Common/GetTypeBasicInformation/Code.py
@@ -20,14 +20,13 @@ MyType = itk.F
 
 print("Min: " + str(itk.NumericTraits[MyType].min()))
 print("Max: " + str(itk.NumericTraits[MyType].max()))
-print("Zero: " + str(itk.NumericTraits[MyType].Zero()))
 print("ZeroValue: " + str(itk.NumericTraits[MyType].ZeroValue()))
 
 print("Is -1 negative? " + str(itk.NumericTraits[MyType].IsNegative(-1)))
 
 print("Is 1 negative? " + str(itk.NumericTraits[MyType].IsNegative(1)))
 
-print("One: " + str(itk.NumericTraits[MyType].One))
+print("OneValue: " + str(itk.NumericTraits[MyType].OneValue()))
 
 print("Epsilon: " + str(itk.NumericTraits[MyType].epsilon()))
 
--- a/src/Core/Common/CreateDerivativeKernel/Code.py
+++ b/src/Core/Common/CreateDerivativeKernel/Code.py
@@ -27,4 +27,4 @@ print("Size: " + str(derivativeOperator.GetSize()))
 print(derivativeOperator)
 
 for i in range(9):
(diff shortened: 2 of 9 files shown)


jhlegarreta:
@mseng10 my bad 🙏 : the bugs I introduced in my recent commits did not allow the CI to shine.

src/Core/Common/GetTypeBasicInformation/Code.py
```diff
@@ -20,7 +20,7 @@
 
 print("Min: " + str(itk.NumericTraits[MyType].min()))
 print("Max: " + str(itk.NumericTraits[MyType].max()))
-print("Zero: " + str(itk.NumericTraits[MyType].Zero()))
+print("Zero: " + str(itk.NumericTraits[MyType].Zero))
```
jhlegarreta, line 23: This is still failing I think. Not sure why `itk.NumericTraits` does not have a `Zero` attribute.

src/Core/common/SetPixelValueInOneImage/Code.py
```diff
@@ -45,7 +45,7 @@
 image = ImageType.New()
 image.SetRegions(region)
 image.Allocate()
-image.FillBuffer(itk.NumericTraits[PixelType].Zero())
+image.FillBuffer(itk.NumericTraits[PixelType].Zero)
```
jhlegarreta, line 48: This should also fail for [the same reason as above](https://github.com/InsightSoftwareConsortium/ITKExamples/pull/174#discussion_r459119051).

jhlegarreta:
The only failures left are those commented in-line.

src/Core/Common/GetTypeBasicInformation/Code.py
thewtex: We need to use `.ZeroValue()` -- we can inspect the functions available with `dir(itk.NumericTraits[MyType])`

jhlegarreta: :+1: Removed the call in 63d0bf6: a call to print the `ZeroValue()` exists immediately after.

src/Core/common/SetPixelValueInOneImage/Code.py
jhlegarreta: Changed to `.ZeroValue()` in 63d0bf6.

jhlegarreta:
As always, thanks for the support @thewtex :100:. It is a pleasure and a great honor to be part of the ITK community !

jhlegarreta:
Hurray ! The failing test on the macOS Cxx build (`MutualInformationAffineTest`) is unrelated to this patch set. So this is ready to be merged.

PR #175, opened 2020-07-22
DOC: Fix the Python code removal line limits

jhlegarreta:
Fix the Python code removal line limits to the appropriate values for each
example.


jhlegarreta:
This is a `DOC` change: the site is not being built yet by the CI and hence test failures are unrelated. Merging.

PR #177, opened 2020-07-24
DOC: Honor the workflows tasks

jhlegarreta:
Honor the workflows tasks: rename the workflow file and name to honor the
tasks being performend: there is no packaging.


jhlegarreta:
Build failures are unrelated (failing Python tests have been addressed in #174). Merging.

PR #193, opened 2021-02-09
ENH: VTK Superbuild Update

mseng10:
## Description
This PR will be the first step in getting the Superbuild's libraries updated to their respective latest versions. In order to avoid a heavy PR, this will strictly contain ITK and VTK changes.

#### Versions Changed
- ITK `abd38d5a0040b9a8fbb0ad3127089dbb72a93342` (same as GitHub Actions)
- VTK `9.0.1`

#### LevelSetV4Visualization 
The examples located in `src/Bridge/VtkGlue` that utilize the (deprecated in ITK `v5.0`) `ITKLevelSetV4Visualization` module, now require the ITK version to be less than `5.0`.

#### VTK Configuration
Several of the examples needed to have their `CMakeLists.txt` reconfigured. In addition, a few of them did not even have VTK utilized in their respective `CMakeLists.txt`, despite utilizing VTK in their `cxx` script. 

After reconfiguring, I am able to get a clean build, but it has resulted in 63 failing tests. However, many of these can be attributed to a single error when using the `VtkGlue` module:
```
7987B]vtkImageSliceMapper.cxx:36    WARN| Error: no override found for 'vtkImageSliceMapper'.
```

I saw a mention of this issue in https://github.com/InsightSoftwareConsortium/ITK/issues/739. I am unsure of the exact steps to fix this, but I will continue to investigate. If anyone has knowledge of this error, please reach out. 

**Note:** The PR will remain a draft until the Superbuild builds cleanly and passes all tests for VTK `9.0.1`.


dzenanz:
I had problems with VTKGlue for a long time, and it is broken in [my nightly build](https://open.cdash.org/viewTest.php?onlyfailed&buildid=7031169).

dzenanz:
Perhaps @mathstuf could provide some advice?

src/Core/Common/FindMaxAndMinInImage/CMakeLists.txt
```diff
@@ -5,9 +5,11 @@ project(FindMaxAndMinInImage)
 find_package(ITK REQUIRED)
 include(${ITK_USE_FILE})
 
+find_package( VTK REQUIRED )
+include_directories(${VTK_INCLUDE_DIRS})
 
 add_executable(FindMaxAndMinInImage Code.cxx)
-target_link_libraries(FindMaxAndMinInImage ${ITK_LIBRARIES})
+target_link_libraries(FindMaxAndMinInImage ${ITK_LIBRARIES} ${VTK_LIBRARIES})
```
mathstuf, line 12: I think you're missing the [`vtk_module_autoinit`](https://vtk.org/doc/nightly/html/group__module.html#gaa19497c3d949b3a93eaa13ed6df5d416) calls for this.

src/Core/Common/FindMaxAndMinInImage/CMakeLists.txt
```diff
@@ -5,9 +5,11 @@ project(FindMaxAndMinInImage)
 find_package(ITK REQUIRED)
 include(${ITK_USE_FILE})
 
+find_package( VTK REQUIRED )
+include_directories(${VTK_INCLUDE_DIRS})
```
mathstuf, line 9: You don't need this anymore; VTK's targets now have the include directories as part of their usage requirements already.

mathstuf (changes requested):
I think you should provide `COMPONENTS` to your `find_package(VTK)` calls. Note that without components and use of `${VTK_LIBRARIES}`, you are linking to *all* of VTK, not just what you need.

mseng10:
After some time, I was able to make significant changes to accommodate examples that utilize VTK/QuickView in some form. In addition to my initial comments in this PR, I was able to successfully configure all of the examples that use VTK and/or QuickView.

For examples that strictly utilize VTK, I was able to configure their respective `CMakeList.txt` to look for the VTK package and only use components that are needed by the example. Then I linked the `VTK_LIBRARIES` and initialized the VTK module using VTK's `vtk_auto_init()`.

For examples that strictly use QuickView, but nothing else, I noticed they weren't being built unless `ENABLE_QUICKVIEW` was set to true. However, all of these examples use the QuickView for viewing purposes and not for the core functionality of the example. Instead, I removed the `if(ENABLE_QUICKVIEW` to enable all of these examples, then in each of their individual `CMakeList.txt`, I enable the VTK libraries it would need if `ENABLE_QUICKVIEW` was set to true. If not set, the example would just build with `ITK_LIBRARIES`. This will greatly improve our test coverage for examples that use ITK by about 46 examples.

Finally, many of the examples needed to have their test input to be updated to properly pass. In addition, their documentation was updated to contain the new input/output. 

A few of the examples and/or their respective tests contained errors
    that were marked with a TODO, for a future PR. These included:
      

1. Baseline comparisons, that cannot work due to the output not ever
         being stored. In a future PR, the examples should be modified to
         save their output, so we can test against it.
            + VTKImageToITKImageBaseline
            + InPlaceFilterOfImageBaseline
2. The DemonstrateThresholdAlgorithm example has it's
         KittlerIllingworth filter commented out with a TODO. This was the
         only filter to not work, and will be documented in a future issue.

thewtex (approved):
@mseng10 fantastic work! :+1:

thewtex:
CC: @wschroed @phcerdan

PR #200, opened 2021-02-22
Small changes to fix missing Code.py files

melrobin:


src/Core/Common/ConvertArrayToImage/CMakeLists.txt
```diff
@@ -14,11 +14,6 @@ install(TARGETS ConvertArrayToImage
   COMPONENT Runtime
   )
 
-install(FILES Code.cxx CMakeLists.txt Code.py
```
thewtex, line 17: We want to keep the block, but remove `Code.py`.

src/Core/Common/ImportPixelBufferIntoAnImage/CMakeLists.txt
```diff
@@ -14,12 +14,6 @@ install(TARGETS ImportPixelBufferIntoAnImage
   COMPONENT Runtime
   )
 
-install(FILES Code.cxx CMakeLists.txt Code.py
-  DESTINATION share/ITKExamples/Code/Core/Common/ImportPixelBufferIntoAnImage/
```
thewtex, line 18: We want to keep the block, but remove `Code.py`.

thewtex (changes requested):
@melrobin thanks for the contribution!

A few comments inline.

thewtex (approved):

PR #217, opened 2021-03-08
BUG: Remove Compatibility/Deprecated group

thewtex:
The ResampleITKVectorImage, ResampleRGBImage are superceded by the
ResampleAVectorImage example.

These examples were incomplete and the only content of
Compatibility/Deprecated. Remove Compatibility/Deprecated because it is
empty. Also, we do not want to instruct folks follow deprecated
practices. This could be added back again if we need to add an example
on how to migrate away from deprecated code.


dzenanz (approved):

jhlegarreta (approved):
Looks fair.

PR #230, opened 2021-03-12
ENH: Add black style pre-commit hook

thewtex:
Black is installed by the Superbuild and run on all changed Python
files during commit. A system version can also be used or configured
with the `black.executable` git configuration value.


src/Segmentation/Watersheds/SegmentWithWatershedImageFilter/Code.py
```diff
@@ -42,6 +42,7 @@
           ' <InputFileName> <OutputFileName> <Threshold> <Level>')
     sys.exit(1)
 
+# aoeuaoe
```
dzenanz, line 48: Mistake?

src/Segmentation/LevelSets/SegmentWithGeodesicActiveContourLevelSet/Code.py
```diff
@@ -21,10 +21,13 @@
     print(
         "Usage: " + sys.argv[0] + "<InputFileName>  <OutputFileName>"
         " <seedX> <seedY> <InitialDistance> <Sigma> <SigmoidAlpha> "
-        "<SigmoidBeta> <PropagationScaling> <NumberOfIterations>")
+        "<SigmoidBeta> <PropagationScaling> <NumberOfIterations>"
```
dzenanz, line 24: Is this file's formatting added to this commit as an example of what black does? Or is it unintentional?

dzenanz (approved):

jhlegarreta (approved):
Apart from Dženan's comment about the unintended comment on ` src/Segmentation/Watersheds/SegmentWithWatershedImageFilter/Code.py`, :+1:

mseng10 (approved):

src/Segmentation/Watersheds/SegmentWithWatershedImageFilter/Code.py
thewtex: Cruft from testing -- good catch :-)

src/Segmentation/LevelSets/SegmentWithGeodesicActiveContourLevelSet/Code.py
thewtex: Yes, this is from black.

Issue #234, opened 2021-03-15
Linting fails on PRs from forked repos


tbirdso:
## Summary
`lint` check passes on local checks in forked repo but fails in PR to dest repo due to annotation permissions.

Example: `nbsphinx` lint check [passes in fork](https://github.com/tbirdso/ITKExamples/runs/2114762761) but [fails in PR](https://github.com/InsightSoftwareConsortium/ITKExamples/pull/213/checks?check_run_id=2114767831).

## Expected behavior
Linting check passes for all valid PRs

## Observed behavior
> Error: This action does not have permission to create annotations on forks. You may want to run it only on `push` events. See https://github.com/wearerequired/lint-action/issues/13 for details
Adding "tbirdso/ITKExamples" fork as remote with Git
Fetching remote branch "nbsphinx"
From https://github.com/tbirdso/ITKExamples
 > * branch              nbsphinx   -> FETCH_HEAD
Switching to the "nbsphinx" branch
>  * [new branch]        nbsphinx   -> fork/nbsphinx
Previous HEAD position was 77bb3abe Merge 47b9936fb990ea891c14d43877723c1fc395ad99 into 66453fdef28d4e4bda2db31925fdc46bfb374b71
Switched to branch 'nbsphinx'
SHA of last commit is "47b9936fb990ea891c14d43877723c1fc395ad99"
Run Black
Create check runs with commit annotations
  Creating GitHub check with 0 annotations for Black…
  Error: Error: Received status code 403
  Error: Error trying to create GitHub check for Black: Received status code 403

tbirdso:
@thewtex did you see this behavior in [6dd94077](https://github.com/InsightSoftwareConsortium/ITKExamples/commit/6dd940777a3a6734c536096558b382faeaea2002)?

thewtex:
@tbirdso thanks for the note -- I had not encountered this, yet.

#235 attempts to address it.

PR #235, opened 2021-03-15
BUG: Use super-linter GitHub Action for black linting

thewtex:
Does not try to create an annotation, which is not permitted on forks
re: https://github.com/wearerequired/lint-action/issues/13

Addresses #234

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,5 +21,6 @@ jobs:
       uses: github/super-linter@v3
       env:
         VALIDATE_ALL_CODEBASE: false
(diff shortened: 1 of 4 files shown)


dzenanz (approved):

.github/workflows/lint.yml
```diff
@@ -21,7 +21,8 @@ jobs:
       run: pip install black
```
tbirdso, line 21: Does `super-linter` require `black` to be manually installed?

thewtex: Let's try without it...

dzenanz (approved):

Issue #238, opened 2021-03-17
Rename ITKExamples repository to ITKSphinxExamples


aylward:
Consider renaming to match the name and directory structure used within ITK

thewtex:
- [ ] Repository renamed on GitHub
- [ ] Update naming references in ITKSphinxExamples repository
- [ ] Update naming references in the ITK
- [ ] Update naming references in the ITKSoftwareGuide repository

PR #244, opened 2021-03-17
STYLE: Rename ITKExamples to ITKSphinxExamples

thewtex:
Re: #238


dzenanz (commented):
Looks good on a glance.

PR #248, opened 2021-03-17
Create tarball

aylward:


aylward:
Oop - need to rebase on master...

dzenanz (commented):
LGTM after rebase.

Utilities/CreateTarball.py
```diff
@@ -24,7 +24,7 @@
 cwd = os.getcwd()
 example_name = sys.argv[1]
 example_dir = os.path.join(cwd, example_name)
-example_idx = example_dir.index('/src/') + 1
+example_idx = example_dir.index('SphinxExamples') + 15
```
thewtex: @aylward thanks for the fix!

When building with the superbuild or externally, though, we will not have *SphinxExamples*.

The original approach could be improved now that we have Python 3. For instance, something like:

```
from pathlib import Path

[...]

cwd = Path(os.getcwd())

example_name = sys.argv[1]
base_dir = cwd.parent.parent.parent
example_relative_path = cwd.relative_to(base_dir) / example_name
html_output_dir = str(Path(sys.argv[2]) / 'html' / example_relative_path)
```

More info: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to

thewtex (approved):

PR #251, opened 2021-03-17
Add MutualInformationAffine Python example file

tbirdso:


dzenanz (commented):
Looks good on a glance. Please squash before marking ready for review.

src/Core/Transform/MutualInformationAffine/Code.py
```diff
@@ -0,0 +1,102 @@
...
+optimizer.SetNumberOfIterations(200)
+optimizer.MaximizeOff()
+
+# Set scales so that the optimizer can take
+# large steps along translation parameters,
+# moderate steps along rotational parameters, and
+# small steps along scale parameters
+optimizer.SetScales([100, 0.5, 0.5, 100, 0.001, 0.001])
+
+registrar = itk.ImageRegistrationMethod[type(fixed_image), type(moving_image)].New()
```
thewtex, line 70: Can we used the functional Python interface here? e.g.

```
result_transform = itk.image_registration_method(fixed_image=fixed_image, moving_image=moving_image, transform=transform, [...]
```

src/Core/Transform/MutualInformationAffine/Code.py
```diff
@@ -0,0 +1,102 @@
...
+    Transform=transform,
+    Input=moving_image,
+    Size=fixed_image.GetLargestPossibleRegion().GetSize(),
+    OutputOrigin=fixed_image.GetOrigin(),
+    OutputSpacing=fixed_image.GetSpacing(),
+    OutputDirection=fixed_image.GetDirection(),
+    DefaultPixelValue=100,
+)
+
+resample.Update()
```
thewtex, line 100: Let's use the simpler, functional interface:

```
resampled = itk.resample_image_filter(moving_image, output_origin=fixed_image.GetOrigin(), [...]
```

thewtex (changes requested):
Nice!

A few requests for the functional interface.

I have not tried the `image_registration_method`, but we should use it, if possible.

src/Core/Transform/MutualInformationAffine/Code.py
tbirdso: The functional Python interface appears to be expecting images of type `signed short` and fails when `unsigned char` images are provided. I've checked that ITK is wrapped for both types. Is there a way to explicitly set the image type? Is this a wrapping that needs to be added to ITK?

Error:
> Traceback (most recent call last):
  File "Code.py", line 73, in <module>
    result_object = itk.image_registration_method(
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\support\itkHelpers.py", line 211, in image_filter_wrapper
    return image_filter(*args, **kwargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\itkImageRegistrationMethodPython.py", line 581, in image_registration_method
    instance = itk.ImageRegistrationMethod.New(*args, **kwargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\support\itkTemplate.py", line 726, in New
    return self[list(keys)[0]].New(*args, **kwargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\itkImageRegistrationMethodPython.py", line 339, in New
    itkTemplate.New(obj, *args, **kargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\support\itkTemplate.py", line 792, in New
    itk.set_inputs(self, args, kargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\support\itkExtras.py", line 928, in set_inputs
    attrib(itk.output(value))
TypeError: in method 'itkImageRegistrationMethodISS2ISS2_SetFixedImage', argument 2 of type 'itkImageSS2 const *'

Verified input types:
> \> print(type(fixed_image))
<class 'itk.itkImagePython.itkImageUC2'>
\> print(type(moving_image))
<class 'itk.itkImagePython.itkImageUC2'>

src/Core/Transform/MutualInformationAffine/Code.py
tbirdso: Same type issue as noted above. Functional interface seems to expect images of type `itk.SS`.
> Traceback (most recent call last):
  File "Code.py", line 93, in <module>
    resampled = itk.resample_image_filter(
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\support\itkHelpers.py", line 211, in image_filter_wrapper
    return image_filter(*args, **kwargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\itkResampleImageFilterPython.py", line 1927, in resample_image_filter
    instance = itk.ResampleImageFilter.New(*args, **kwargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\support\itkTemplate.py", line 726, in New
    return self[list(keys)[0]].New(*args, **kwargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\itkResampleImageFilterPython.py", line 680, in New
    itkTemplate.New(obj, *args, **kargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\support\itkTemplate.py", line 792, in New
    itk.set_inputs(self, args, kargs)
  File "C:\repos\ITK-builds-2\Wrapping\Generators\Python\itk\support\itkExtras.py", line 928, in set_inputs
    attrib(itk.output(value))
TypeError: Expecting argument of type itkImageSS2 or itkImageSourceISS2.
Additional information:
Wrong number or type of arguments for overloaded function 'itkImageToImageFilterISS2ISS2_SetInput'.
  Possible C/C++ prototypes are:
    itkImageToImageFilterISS2ISS2::SetInput(itkImageSS2 const *)
    itkImageToImageFilterISS2ISS2::SetInput(unsigned int,itkImageSS2 const *)

src/Core/Transform/MutualInformationAffine/Code.py
tbirdso: Also note that `itk.image_registration_method` returns an object of type `'itk.itkTransformBasePython.itkDataObjectDecoratorTD22'` which cannot be passed directly to `resample_image_filter` as an `itk.transform`. Not a big deal given that the registration transform is updated in-place and can be passed on, but I'm not sure what this object contains. Where can I find documentation for the functional interface?

thewtex: We can be explicit with the types we want with the [`ttype` keyword argument](https://github.com/InsightSoftwareConsortium/ITK/blob/2880b97c88100ddaf809204d6a156bfb5e694e85/Wrapping/Generators/Python/itk/support/itkTemplate.py#L678).

However, we could have the filter select the proper default template type by looking at the first positional argument / template argument. This corresponds to the first indexed input to the filter, the primary input. While `itk::ImageRegistrationMethod` does not identify the `FixedImage` as its primary input, `itk::ImageRegistrationMethodv4` does:

https://github.com/InsightSoftwareConsortium/ITK/blob/cde72b3e7d716542b11d2dbb402bbe5816cf731f/Modules/Registration/RegistrationMethodsv4/include/itkImageRegistrationMethodv4.hxx#L45

So, we should update this examples to use the v4 registration method.

src/Core/Transform/MutualInformationAffine/Code.py
thewtex: What is `type(moving_image)`?

tbirdso: > \> print(type(moving_image))
<class 'itk.itkImagePython.itkImageUC2'>

Note that no error is thrown when reading in images of type itk.SS but `itk.imwrite` does not support writing `itk.SS` PNG output images by default.

Per above, this may be resolved by updating to the v4 metric.

src/Core/Transform/MutualInformationAffine/Code.py
tbirdso: This example currently relies on the Viola-Wells mutual information metric but it seems that only the Mattes mutual information metric has a v4 counterpart. Given that a v4 metric is required for the v4 registration method, is it suitable to explicitly set `ttype` in this example for the v3 metric and open an issue to add a new example using the Mattes mutual information v4 metric? It doesn't look like we currently have an example making use of `MattesMutualInformationImageToImageMetricv4`.

hjmjohnson:
Closing as stale after 5 years without updates. Adding a Python example for `MutualInformationAffine` is still a good idea — if anyone wants to revive this, please rebase on the current `main` branch and open a fresh PR.

Thank you @tbirdso for the contribution.

PR #253, opened 2021-03-17
ENH: Use the upstream CMake ExternalData module, instead of a vendored local copy

brianhelba:
The minimum supported CMake version should now always contain ExternalData.


dzenanz (approved):
Commit message could use `ENH: ` prefix, otherwise great!

brianhelba:
@dzenanz Done.

dzenanz (approved):

thewtex (approved):
Awesome @brianhelba !

PR #254, opened 2021-03-17
ENH: `ScaleAnImage` Python example and input data update

tbirdso:
- Adds Python example and baseline test
- Replaces contrived square image input generated at runtime with static `cthead1.png` input

Input

ScaleAnImageInput

Output

ScaleAnImageOutputBaseline


dzenanz (approved):

src/Core/Transform/ScaleAnImage/Code.py
```diff
@@ -0,0 +1,42 @@
...
+
+import sys
+
+import itk
+
+if len(sys.argv) < 3:
+    raise Exception(f"Usage: {sys.argv[0]} inputFile outputFile")
+
+image = itk.imread(sys.argv[1], itk.UC)
+
```
thewtex, line 29: `Dimension = image.GetImageDimension()`, then use this in the `ScaleTransform` initialization to make the script dimension-agnostic.

src/Core/Transform/ScaleAnImage/Code.py
```diff
@@ -0,0 +1,42 @@
...
+transform = itk.ScaleTransform[itk.D, 2].New(
+    Scale=[1.5,1.5])
+
+center = [x / 2 for x in list(image.GetLargestPossibleRegion().GetSize())]
```
thewtex, line 33: `list(image.GetLargestPossibleRegion().GetSize()` -> `itk.size(image)`

src/Core/Transform/ScaleAnImage/Code.py
```diff
@@ -0,0 +1,42 @@
...
+transform.SetCenter(center)
+
+filter = itk.ResampleImageFilter[type(image), type(image)].New(
+    Transform=transform,
+    Input=image,
+    Size=image.GetLargestPossibleRegion().GetSize())
+filter.Update()
```
thewtex, line 40: -> `scaled = itk.resample_image_filter(image, transform=transform, size=itk.size(image))`

thewtex (changes requested):
Nice! A few Pythonic improvements noted inline.

tbirdso:
Seeing a difference in baseline comparisons between cxx and Python examples despite attempting to run the exact same set of procedures for image scaling. Cxx output matches baseline (which was generated with the cxx test) and appears visually correct, Python output does not appear to scale the image border as expected.

I've tested both `itk.ResampleImageFilter` class directly and the `itk.resample_image_filter` functional interface, with the same results. @dzenanz do you have any thoughts on possible reasons why we might get different output from similar input to `itk::ResampleImageFilter`?

Input
ScaleAnImageInput

Cxx output
ScaleAnImageOutput

Python output
ScaleAnImageOutputPython

src/Core/Transform/ScaleAnImage/Code.py
```diff
@@ -0,0 +1,40 @@
...
+
+if len(sys.argv) < 3:
+    raise Exception(f"Usage: {sys.argv[0]} inputFile outputFile")
+
+image = itk.imread(sys.argv[1], itk.UC)
+
+Dimension = image.GetImageDimension()
+
+transform = itk.ScaleTransform[itk.D, 2].New(
+    Scale=[1.5] * Dimension, Center=[x / 2 for x in itk.size(image)]
```
dzenanz, line 33: In C++, center is explicitly `float`. Here it might be implicitly `double`.

dzenanz (commented):
If float/double difference is not it, I don't know what might be causing it.

src/Core/Transform/ScaleAnImage/Code.py
thewtex: Does changing `x / 2` to `x / 2.` address the issue?

tbirdso: Unfortunately casting with `x / 2.` or `float(x / 2)` does not change the image output. I also tried explicitly setting `center` to `double` in the cxx test but it didn't seem to change the cxx image output either. In both cases the cxx baseline comparison test continues to succeed while the python baseline comparison test fails.

tbirdso:
Closing this stale PR as I haven't been able to pinpoint the issue source and the issue itself seems trivial as it is a difference of one pixel width. Unfortunately there isn't a good way to reference a different baseline image for Python vs cxx tests here.

If this is seen as a symptom of a larger problem (Python vs cxx behavior difference) we can re-submit as an issue against ITK.

dzenanz:
An open PR is visible, and people can "steal" code from it easier. If we don't have need/time to resolve the issue now, we should keep it open until we do, or until someone else suggests a fix.

tbirdso:
Is it possible this is related to [ITK Issue 2703](https://github.com/InsightSoftwareConsortium/ITK/issues/2703)?

Note that the root of discussion there is [ResampleInPlaceFilter](https://discourse.itk.org/t/simpleitk-euler3d-transform-adjust-image-origin-and-direction/3966/4) [(docs)](https://itk.org/Doxygen/html/classitk_1_1ResampleInPlaceImageFilter.html) and focuses on image direction discrepancies while this example uses [ResampleImageFilter](https://itk.org/Doxygen/html/classitk_1_1ResampleImageFilter.html) and discrepancies in pixel data in the output image. It does not appear that these two filters inherit from each other.

tbirdso:
Pinging @brad-t-moore if he is available to get another set of eyes on this issue

dzenanz:
I don't think this is related to issue [2703](https://github.com/InsightSoftwareConsortium/ITK/issues/2703).

src/Core/Transform/ScaleAnImage/Code.py
tbirdso: Updating to read in input data rather than generate input in C++ and Python tests respectively seems to have resolved the issue.

tbirdso:
@thewtex Ping for re-review

src/Core/Transform/ScaleAnImage/Code.py
```diff
@@ -0,0 +1,40 @@
...
+)
+
+scaled_image = itk.resample_image_filter(
+    Input=image, Transform=transform, Size=itk.size(image)
```
thewtex, line 37: Nitpick: `Input` -> `input`, `Transform` -> `transform`, `Size` -> `size`

thewtex (approved):
Looks great!

Minor style nitpick inline

src/Core/Transform/ScaleAnImage/Code.py
tbirdso: Thanks, fixed!

tbirdso:
Resolved merge conflicts.

PR #256, opened 2021-03-18
DOC: Rename build-test-superbuild to build-test-python

thewtex:
Contrast and identify these builds with build-test-cxx and
build-test-notebooks: these builds use the superbuild, but in the
process they test the Python examples in addition to the C++ examples.

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -125,7 +125,7 @@ jobs:
           ctest -j 2 -VV -S dashboard.cmake
         shell: cmd
 
-  build-test-superbuild:
+  build-test-python-superbuild:
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-publish.yml
```diff
@@ -125,7 +125,7 @@ jobs:
           ctest -j 2 -VV -S dashboard.cmake
         shell: cmd
 
-  build-test-superbuild:
+  build-test-python:
```
tbirdso, line 128: If these checks are intended to cover both superbuild and Python example functionality, could we use a name indicative of that intention such as `superbuild-test-python`?

thewtex: Yes, we could use `python-superbuild`.

mseng10 (approved):

tbirdso (approved):

PR #257, opened 2021-03-18
DOC: Put remaining example headers into proper format

mseng10:
Several of the examples needed their headers put into proper MLA/APA format. In addition, it ensures `CreateAnRGBImage/Documentation.rst` is in proper alphabetical order when being displayed in Sphinx.


dzenanz (approved):

thewtex (approved):
:100:

PR #268, opened 2021-03-24
Fix another small doc file

melrobin:
Another small doc fix.  Allows make install to complete when BUILD_EXAMPLES=ON in ITK.


dzenanz (approved):

thewtex (approved):
Thanks, @melrobin !

PR #275, opened 2021-05-06
ENH: Update ITK to v5.2.0

thewtex:


Superbuild/External-Python.cmake
```diff
@@ -17,6 +17,6 @@ ExternalProject_Add(ITKPython
   DOWNLOAD_COMMAND ""
   CONFIGURE_COMMAND ${PYTHON_EXECUTABLE} -m venv "${_itk_venv}"
   BUILD_COMMAND ${ITKPYTHON_EXECUTABLE} -m pip install --upgrade pip
-  INSTALL_COMMAND ${ITKPYTHON_EXECUTABLE} -m pip install --ignore-installed itk>=5.2rc3 sphinx==3.0.4 six black nbsphinx
+  INSTALL_COMMAND ${ITKPYTHON_EXECUTABLE} -m pip install --ignore-installed itk>=5.2.0.post2 sphinx==3.0.4 six black nbsphinx
```
jhlegarreta, line 20: Should this be `5.2.0` (without the `.post2`)?

thewtex: The `.post2` is needed re: https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/pull/273 -- `.post2` removes the need for that workaround. After integration here, we can revert that patch.

tbirdso (approved):

jhlegarreta (approved):

dzenanz (approved):

Superbuild/External-Python.cmake
Leengit: To supply a little more detail: ITK commit [ca069006](https://github.com/InsightSoftwareConsortium/ITK/pull/2524/commits/ca0690069a407b32b4e1f5232c8d36ec24a6ac92) is the commit that fixes the `ellipsis is not iterable` error.  That commit is newer than the `v5.2.0` tag; hence the need for the later `v5.2.0.post2` tag.  However, that is a Python-only fix, and the earlier `v5.2.0` tag works for ITK C++ code.

Leengit (approved):

PR #277, opened 2021-05-14
geodesic notebook

thewtex:
- STYLE: GeodesicActiveContour CMake style
- DOC: Add notebook example for SegmentWithGeodesicActiveContourLevelSet

PR #278, opened 2021-05-14
Switch to the new code order: Python before C++

dzenanz:
I think the PR which introduced this example was concurrent to the one which swapped the order.


thewtex (approved):
:+1:

thewtex:
Notebook CI issues should be addressed by #277

PR #290, opened 2021-05-28
ENH: Added ReadAnImage Python script

kian-weimer:
Created a Python ReadAnImage example based off of the existing C++ example.

### DIFF
--- a/src/IO/ImageBase/ReadAnImage/Code.py
+++ b/src/IO/ImageBase/ReadAnImage/Code.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python
+
+# Copyright NumFOCUS
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


kian-weimer:
Would it be preferable to have multiple examples in one PR?

dzenanz (approved):
I am fine with one PR per example. Others might have a different preference, but it does not matter much. If you do that, please have each new example in a commit of its own.

dzenanz (approved):

dzenanz:
Could you rebase on master? The log should be shorter then.

kian-weimer:
Rebased. Thanks for the tip. I'm not sure what the issue is yet.

src/IO/ImageBase/ReadAnImage/Code.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
```
thewtex, line 1: The license header is missing

thewtex (dismissed):
Please copy the license header from another Python example.

src/IO/ImageBase/ReadAnImage/Code.py
kian-weimer: Sorry I missed this! 
What's the preferred method for fixing this on my other already committed PRs?

dzenanz: Make a new PR, which adds the missing headers and modifies the corresponding .rst to skip including the licenses.

kian-weimer: Thank you, will do.

dzenanz (approved):

PR #297, opened 2021-06-06
ENH: Added LaplacianRecursiveGaussianImageFilter Python script

kian-weimer:
Created a Python LaplacianRecursiveGaussianImageFilter example based off of the existing C++ example.

### DIFF
--- a/src/Filtering/ImageFeature/LaplacianRecursiveGaussianImageFilter/Code.py
+++ b/src/Filtering/ImageFeature/LaplacianRecursiveGaussianImageFilter/Code.py
@@ -1,0 +1,28 @@
+#!/usr/bin/env python
+
+# Copyright NumFOCUS
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import itk
+
+if len(sys.argv) != 3:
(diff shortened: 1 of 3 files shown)


src/Filtering/ImageFeature/LaplacianRecursiveGaussianImageFilter/Code.py
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import itk
+
+if len(sys.argv) != 3:
```
dzenanz, line 20: Since you are creating many new examples, maybe you could use [CreateNewExample.py](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/blob/master/Utilities/CreateNewExample.py.in) (which appears in the build directory) to start new examples, so they use the argument parser from the template:

https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/blob/22f432f69041fd65ce96137915b1ef4e664213a9/Utilities/CookieCutter/%7B%7Bcookiecutter.example_name%7D%7D/Code.py#L20-L23

dzenanz (approved):

src/Filtering/ImageFeature/LaplacianRecursiveGaussianImageFilter/Code.py
kian-weimer: Sure, I'm just adding Python versions of existing examples, but I can follow that portion of the template. I can change the remaining PRs to fit this. The example that is created seems to be missing an import for `argparse`. Is this indirectly imported or should I be adding an import? 
I noticed that the cookie cutter example had the Python Doc after the C++. It is preferred to have the Python code show up before the C++ in the documentation correct?

dzenanz: > change the remaining PRs

Please do! You can also update some existing examples to this style in a separate PR.

> adding an import

If it is needed, yes 😄

> It is preferred to have the Python code show up before the C++ in the documentation

Correct.

PR #304, opened 2021-07-07
ENH: Add Python version of RegisterImageToAnotherUsingLandmarks

thewtex:
A few tweaks also to the C++ version.

### DIFF
--- a/src/Registration/Common/RegisterImageToAnotherUsingLandmarks/CMakeLists.txt
+++ b/src/Registration/Common/RegisterImageToAnotherUsingLandmarks/CMakeLists.txt
@@ -14,7 +14,7 @@ install(TARGETS RegisterImageToAnotherUsingLandmarks
   COMPONENT Runtime
   )
 
-install(FILES Code.cxx CMakeLists.txt
+install(FILES Code.cxx CMakeLists.txt Code.py
   DESTINATION share/ITKSphinxExamples/Code/Registration/Common/RegisterImageToAnotherUsingLandmarks/
   COMPONENT Code
   )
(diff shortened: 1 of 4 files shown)


src/Registration/Common/RegisterImageToAnotherUsingLandmarks/moving.png.sha512
```diff
@@ -1 +1 @@
-9411c2f629a4c1d833b517c85527e85990d85d4e5f45edc52f5b5ab6b14ff1ffa54ad7b28ce56023456a4602c25e49b2acb81d216b4620efde3b615655f4b474
+376682f715f8d870b1bf91ae037914447220b5d8a11546daaf0df0fe1eec7bc35f5a53f563287849740abb7fe74b4b4c002b1c4d1be646acc614bb4b018a9e82
```
dzenanz, line 1: Why do the images need to be updated?

dzenanz (approved):

dzenanz (approved):

src/Registration/Common/RegisterImageToAnotherUsingLandmarks/moving.png.sha512
thewtex: These snuck in there somehow. Now removed.

jhlegarreta (approved):
Have not tried but looks good. A couple of minor comments:
- An `ENH` commit prefix would honor better the contribution.
- I think we have not done so in ITK tests either, and could be definitely be done in a future PR, but I'm wondering whether the creation of the test fixed and moving images could be put into some `utils` module so that they can be used.

Issue #226, opened 2021-03-11
Use itk::ReadImage and itk::WriteImage 


thewtex:
Use the new functions, `itk::ReadImage` and `itk::WriteImage`, available in ITK 5.2 for simplicity over `itk::ImageFileReader` and `itk::ImageFileWriter` throughout the examples. See https://github.com/InsightSoftwareConsortium/ITKExamples/pull/225
for an illustration of the changes required.

josetascon:
Hi @thewtex,
I would like to contribute. I have a question. Is this desired for all the examples?. If that is the case, should I make individual commits for every example?

For instance, If I check how many example files uses itkImageFileReader there are 212, while itkImageFileWriter 210. It may be convenient to commit per module (Core, Numerics, ...)

I will make a single commit with the example at src/Core/Transform/ScaleAnImage/Code.cxx

dzenanz:
> commit per module

This sounds good. Or change all in one commit. There would be too many commits with one commit per example. #302 is just what we are looking for. This should be done for great majority of the examples. If reader or writer are being used to demonstrate streaming or in some other unusual way, they should be left alone.

josetascon:
Ok. New commit #303 with changes of examples at src/Core. 
I have some notes. Some of comments are just to mention the unchanged, another file changed, some examples without test, and finally some examples that import header files but they are unnecessary.

Unchaged
    * ImportPixelBufferIntoAnImage unchanged because demonstrate filter connection
    * MiniPipeline unchanged because demonstrate itk pipeline
    * StoreNonPixelDataInImage unchanged because reader calls reader->GetMetaDataDictionary()
    * MakePartOfImageTransparent unchanged because sets extra writer parameters
    * TryCatchException unchanged because use the reader to catch an exception

Changed
    * ConvertArrayToImage change output format to mhd

Missing compiled binary (not added in cmake) and without test files
    * MultipleInputsOfDifferentType 
    * MultipleOutputsOfDifferentType
    * FilterImageUsingMultipleThreads
    * IterateRegionWithNeighborhood
    * MakeOutOfBoundsPixelsReturnConstValue
    * GetOrSetMemberVariableOfITKClass
    * InPlaceFilterOfImage

Not working as intended, all test passed but the output is a black image
    * ContourSpatialObject
    * CreateALineSpatialObject

Unused header
    Import file reader and writer headers but are not used:
    * FilterImage
[...]

PR #305, opened 2021-07-14
STYLE: Use itk::ReadImage and itk::WriteImage in src/Filtering examples

josetascon:
Refs #226


dzenanz (approved):
Thousand of lines removed, more than 6 per file on average!

PR #306, opened 2021-07-21
STYLE: Use itk::ReadImage and itk::WriteImage in src/ Reg., Seg., Num.

josetascon:
Refs #226 in modules src/Registration, src/Segmentation, src/Numerics


dzenanz (approved):
Great! Thank you for the effort.

PR #310, opened 2021-09-29
ENH: Add labeler workflow

jhlegarreta:
Add labeler workflow so that a `GitHub` action, triggered when a PR is
opened, automatically labels the PR based on the given rules.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,96 @@
+# Commit prefix regexes
+type:Compiler:
+  title: "^COMP:.*"
+
+type:Bug:
+  title: "^BUG:.*"
+
+area:Documentation:
+  title: "^DOC:.*"
+
+type:Enhancement:
+  title: "^ENH:.*"
+
+type:Performance:
+  title: "^PERF:.*"
+
+type:Style:
+  title: "^STYLE:.*"
+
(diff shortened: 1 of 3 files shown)


jhlegarreta:
Took the freedom to import the relevant labels from the `ITK` repository:
 - Added some examples-specific labels: `Python Notebook`, `language:C++`, `language:Python`
 - Removed: `action:ApplyClangFormat`, `area:Examples`, `area:Python wrapping`
 
Additionally, the following labels:
 - Added a `Nonunit` type
 - Added a `Data` type.
 
Might be worthwhile to be added to ITK as well. I ignore why they were not added back at the time. Let me know what you think.

jhlegarreta:
I excluded the `labeler.yml` file from the linter because the linter would like it to have some heading, which I think is not necessary, and found a duplication for `area:Documentation` which is not strictly correct since one should operate on commit prefixes and the other one on filenames:
https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/runs/3750344302?check_suite_focus=true#step:7:74

jhlegarreta:
I guess we will not know if it works properly until it gets merged, much like the one in ITK did.

dzenanz (approved):

Issue #308, opened 2021-09-17
CreateImage does not clear the buffer


dzenanz:
A search for `image->Allocate();` has 149 hits while `image->FillBuffer(` has 64 hits. The instances of `Allocate` not followed by a `FillBuffer` should be examined, and probably turned into [image->Allocate(true);](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/commit/fc491a42151e2dd27f3e98b821d0c2fd35b55b2b#diff-a388a43bc1ceb769184a4daea4a146c73cbe1c5aeaf7003d1c698f51f95f0ad6R113) calls.

PR #312, opened 2021-11-29
BUG: Initialize image buffer in Inverse FFT example

tbirdso:
Previous behavior: FFT and baseline comparison sometimes returns NAN pixel values due to uninitialized buffer

Updated behavior: Pixels are initialized to 0; FFT and baseline comparison succeed


dzenanz (approved):
See also #308.

tbirdso:
Docs failure seems unrelated to this change: [https://open.cdash.org/viewBuildError.php?type=1&buildid=7597891](https://open.cdash.org/viewBuildError.php?type=1&buildid=7597891)

> WARNING: while setting up extension sphinx.addnodes: node class 'meta' is already registered, its visitors will be overridden

PR #313, opened 2021-12-09
BUG: Corrected VTK image from image conversion. 

kian-weimer:
Switched incorrect 'image_to_vtk_image' function call to the correct 'vtk_image_from_image'.


dzenanz:
The [warning](https://open.cdash.org/viewBuildError.php?type=1&buildid=7616221) `WARNING: while setting up extension sphinx.addnodes: node class 'meta' is already registered, its visitors will be overridden` seems unrelated to this PR.

PR #314, opened 2021-12-17
remove recursive header includes

hjmjohnson:
COMP: Remove inclusion of .hxx files as headers

The ability to include either .h or .hxx files as
header files required recursively reading the
.h files twice.  The added complexity is
unnecessary, costly, and can confuse static
analysis tools that monitor header guardes (due
to reaching the maximum depth of recursion
limits for nested #ifdefs in checking).

### DIFF
--- a/src/Developer/ImageFilter.hxx
+++ b/src/Developer/ImageFilter.hxx
@@ -1,7 +1,6 @@
 #ifndef __itkImageFilter_hxx
 #define __itkImageFilter_hxx
 
-#include "ImageFilter.h"
 #include "itkObjectFactory.h"
 #include "itkImageRegionIterator.h"
 #include "itkImageRegionConstIterator.h"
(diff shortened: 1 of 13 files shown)


src/Developer/ImageFilter.hxx
```diff
@@ -1,7 +1,6 @@
 #ifndef __itkImageFilter_hxx
 #define __itkImageFilter_hxx
 
-#include "ImageFilter.h"
```
dzenanz, line 4: Is a similar change needed in the software guide? If there are any instances of this, it is.

dzenanz (approved):

src/Developer/ImageFilter.hxx
jhlegarreta: https://github.com/InsightSoftwareConsortium/ITKSoftwareGuide/pull/171 for the coding style guidelines section.

PR #316, opened 2021-12-22
ENH: Miscellaneous fixes to developer files

jhlegarreta:
- BUG: Include the appropriate implementation file 
- BUG: Fix the class name in the `ImageSource` source files 
- BUG: Inherit `ImageSource` from ProcessObject`
- BUG: Make the include guards match the filenames 
- STYLE: Do not use double underscore defines 
- COMP: Use and move ITK_DISALLOW_COPY_AND_MOVE calls to public section
- STYLE: Move default construtor to .h
- STYLE: Prefer = default to explicitly trivial implementations
- COMP: Mark class destructors with `override`
- STYLE: Use typename instead of class in template definitions 
- ENH: Add `const` smart pointer type alias

### DIFF
--- a/src/Developer/ImageFilter.h
+++ b/src/Developer/ImageFilter.h
@@ -1,5 +1,5 @@
-#ifndef __ImageFilter_h
-#define __ImageFilter_h
+#ifndef ImageFilter_h
+#define ImageFilter_h
 
 #include "itkImageToImageFilter.h"
 
@@ -37,4 +37,4 @@ protected:
 #endif
 
 
-#endif // __ImageFilter_h
+#endif // ImageFilter_h
--- a/src/Developer/ImageFilter.h
+++ b/src/Developer/ImageFilter.h
@@ -14,6 +14,7 @@ public:
   using Self = ImageFilter;
   using Superclass = ImageToImageFilter<TImage, TImage>;
   using Pointer = SmartPointer<Self>;
+  using ConstPointer = SmartPointer<const Self>;
 
   /** Method for creation through the object factory. */
   itkNewMacro(Self);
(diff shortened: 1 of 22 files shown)


jhlegarreta:
Not sure if these fix all the`Warning Fix Errors Example contains errors needed to be fixed for proper output.` notice in the related pages, probably takes care of a few of them:
https://itk.org/ITKExamples/src/Core/Common/ProduceImageProgrammatically/Documentation.html

Left for discussion and a separate PR:
- Should the classes in the `ImageFilterX.*` and `ImageFilterY.*` be renamed to `ImageFilterX` and `ImageFilterY` ?
- Should all files be renamed to have a leading `itk*`? There are actually some inconsistencies in classes with supposedly similar patterns, e.g. `itkImageFilterMultipleInputs.*` vs `ImageFilterMultipleOutputs`.*` -not sure if intended). This will need to be accompanied with the  appropriate changes to the include guards.

jhlegarreta:
Not sure whether the Python linter has beein failing lately due to some library not being found:
https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/runs/4602491214?check_suite_focus=true#step:7:11

But this PR does not modify any Python file.

jhlegarreta:
Reported failures were already present in previous PRs. Might be related the required C++ version in the ITK Examples or its CI builds being lower to the one in ITK:
https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/runs/4602491312?check_suite_focus=true#step:11:517
or
https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/runs/4602491312?check_suite_focus=true#step:11:528

dzenanz (approved):
Thanks for this update and cleanup!

PR #317, opened 2021-12-27
Fix CI errors

dzenanz:
I am not sure whether this will fix all errors, but it will surely fix some.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,7 +18,7 @@ jobs:
         python-version: 3.9
 
     - name: Lint Python
-      uses: github/super-linter@v3
+      uses: github/super-linter@v4
       env:
         VALIDATE_ALL_CODEBASE: false
         VALIDATE_PYTHON_BLACK: true
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,15 +35,15 @@ foreach(pold "") # Currently Empty
   endif()
 endforeach()
 
-# ==== Define language standard configurations requiring at least c++11 standard
-if(CMAKE_CXX_STANDARD EQUAL "98" )
-  message(FATAL_ERROR "CMAKE_CXX_STANDARD:STRING=98 is not supported in ITK version 5 and greater.")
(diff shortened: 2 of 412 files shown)


hjmjohnson (approved):

hjmjohnson (approved):

hjmjohnson (approved):

dzenanz:
After this PR, we are just [one warning](https://open.cdash.org/viewBuildError.php?type=1&buildid=7646735) short of a clean CI:
```text
[1587/1588] Generating sphinx documentation: html
Running Sphinx v3.0.4
WARNING: while setting up extension sphinx.addnodes: node class 'meta' is already registered, its visitors will be overridden
```

jhlegarreta:
> After this PR, we are just one warning short of a clean CI:

Awesome @dzenanz !

Looks like the error stems from using `docutils` 0.18.1 (latest) and a not very recent `sphinx` (latest is 4.3.2):
https://github.com/sphinx-doc/sphinx/issues/9841#issuecomment-971739437

The CI's `sphinx` version is set in:
https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/blob/master/.github/workflows/build-test-publish.yml#L285

I'd say that the `sphinx` version be updated. @thewtex is there a particular, more recent version that we should adopt?

dzenanz:
Matt is currently out of office, and is generally busy. Maybe we should try 3.5.4 (released Apr 11, 2021), as the latest from 3.x branch? Or the latest, 4.3.2 (released Dec 19, 2021)? I guess that we should go with the latest one which works for us.

jhlegarreta:
I asked because maybe the version should match that one used to actually build the site, and I do not know whether it was kept at 3.0.4 for some particular reason. I'm fine if you decide that we should go either with the latest from 3.x or the most recent one.

dzenanz:
I think that Sphinx documentation is built and deployed upon merge. Or at least it was. So any configuration should be in the repository (CI etc).

dzenanz:
Going with 3.x is less likely to cause problems (e.g. new warnings), so I think that is lower hanging fruit.

PR #320, opened 2022-01-17
COMP: ENABLE_QUICKVIEW code updates

hjmjohnson:
Conditional code for ENABLE_QUICKVIEW no
longer compiled.

Changes to use simpler "auto image = ReadImage<>()"
function changed the signature of the variable type
for image.


hjmjohnson:
Lint failure seems to be a problem with the docker image missing a file.

dzenanz (approved):

thewtex (approved):
Thanks, @hjmjohnson !

PR #327, opened 2022-03-13
BUG: Ignore breathe duplicate C++ warnings in docs CI

### LINKED ISSUE #266: Duplicate declaration warnings on dashboard
The `Linux Examples Documentation` build of the Insight dashboard:
https://open.cdash.org/viewBuildError.php?type=1&buildid=7112107

is signalling duplicate declaration warnings for the classes that are marked as being demonstrated in each example's `Documentation.rst` file with the `.. breathelink:: ` tag:
e.g.
```
[CTest: warning matched]
/home/kitware/Dashboards/Tests/ITKExamples-doc-build/ITKEx-buildCore/Common/
BuildAHelloWorldProgram/Documentation.rst:102:
WARNING: Duplicate declaration,
template<typename TPixel, unsigned int VImageDimension = 2> itk::Image :
public itk::ImageBase< VImageDimension >
```

Not sure why this is being reported. Also, the build seems not to trigger nightly (or it might take a long time to build), and hence the reporting frequency is not consistent.

tbirdso:
Resolves #266

### DIFF
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -341,6 +341,8 @@ jobs:
             \${CTEST_CUSTOM_WARNING_EXCEPTION}
             # macOS Azure VM Warning
             "ld: warning: text-based stub file"
+            # breathelink warning
+            "WARNING: Duplicate C"
             )
           set(dashboard_no_clean 1)
           set(ENV{CC} ${{ matrix.c-compiler }})
@@ -369,8 +371,8 @@ jobs:
         with:
           name: Documentation
           path: |
-            ${{ github.workspace }}/../../bld/ITKEx-build/ITKSphinxExamples-*-html.tar.gz
-            ${{ github.workspace }}/../../bld/ITKEx-build/ITKSphinxExamples-*-html.zip
+            /home/runner/work/bld/ITKEx-build/ITKSphinxExamples-*-html.tar.gz
+            /home/runner/work/bld/ITKEx-build/ITKSphinxExamples-*-html.zip
 
       - name: Unpack site
         shell: bash
--- a/.github/workflows/build-test-publish.yml
+++ b/.github/workflows/build-test-publish.yml
@@ -378,7 +378,7 @@ jobs:
         shell: bash
         run: |
           mkdir site
-          tar --strip-components=1 -xf ${{ github.workspace }}/../../ITKEx-build/ITKSphinxExamples-*-html.tar.gz -C site
+          tar --strip-components=1 -xf /home/runner/work/bld/ITKEx-build/ITKSphinxExamples-*-html.tar.gz -C site
 
       - name: Publish website to Netlify
         uses: nwtgck/actions-netlify@v1.1.13


dzenanz:
As far as I can tell, the current error is:
```text
Cloning into 'zlib'...
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```
quick fix is to use `https://` protocol for fetching.

tbirdso:
Thanks @dzenanz , I pushed a change this morning removing most warning exceptions from `CTestCustom.cmake.in` to try to verify whether changes to the file are having an impact. Unfortunately I can't tell whether this error is actually related 🤔 I am going to try re-running CI and see if the error persists.

tbirdso:
Observed the same warnings on re-run, so _something_ is happening with `CTestCustom.cmake`, but I'm still not sure why the regular expressions for excluding the duplication warning are not being picked up. Reverting so that other warning exceptions are populated.

tbirdso:
The `zlib` error is present in main branch CI after rerun. I will enter a new issue and PR to track. @dzenanz as you said this should be fixable by forcing `https` rather than `git` to be used for retrieving files.

thewtex (approved):

tbirdso:
Rebased on `master` to fix zlib warnings.

tbirdso:
The CI re-run did not capture updated changes. It looks like Github Actions is having an outage, may be best to wait for coverage to resume before attempting further changes. [https://www.githubstatus.com/](https://www.githubstatus.com/)

tbirdso:
From what I can tell `CTestCustom.cmake` is having no impact on CI.

After removing most warning exceptions, decreasing the number of warnings and errors, and adding additional regex expressions to except, [CDash results](https://open.cdash.org/viewBuildError.php?type=1&buildid=7802896) are unchanged at 50 warnings about duplicate C++ declarations.

@thewtex @dzenanz Do you have thoughts on steps we can take for debugging here? As this file is expected to solely affect CDash behavior I am a bit limited in what testing I can do.

dzenanz:
@hjmjohnson recently made some docs-related PR (https://github.com/InsightSoftwareConsortium/ITK/pull/3311), he might have some suggestions. I have not dealt with docs building in years (maybe ever). I would have to dive in, just like you.

tbirdso:
Aha, new findings:
- The `CTEST_CUSTOM_WARNING_EXCEPTION` variable is extended in the docs CI procedure when we create `dashboard.cmake` to drive the CTest dashboard procedure. We can resolve #266 by editing `build-test-publish.yml` accordingly.
- [CI](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/runs/5612796881?check_suite_focus=true) failed at the publish-artifact stage due to relative pathing. Perhaps the absolute pathing requirement is recent? We can work around it by using the absolute path to the work directory.
- My current guess as to why CTestCustom.cmake is not affecting dashboard results is that the file is not moved to the right location as part of the superbuild procedure. After running the CTest command locally I see that `CTestCustom.cmake` is configured in the `ITKEx-build` subfolder of the superbuild, while it should be at the top level directory. Given that the build process has not actually been using it, perhaps it would be prudent to simply delete it? Alternatively we can move it to the top level in the CI procedure.

dzenanz (approved):

PR #334, opened 2022-03-24
ENH: Deploy docs to Fleek

tbirdso:
Moving docs deployment to [Fleek](https://fleek.co/)


tbirdso:
These checks look different than other PRs. Surprised to see a `Netlify` check still running despite this being a branch on the original repo. Will these CI changes not be applied until merge?

Issue #335, opened 2022-03-25
HTML build fails with `jinja2` warning


tbirdso:
Raw build log from [docs CI](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/runs/5680532102?check_suite_focus=true) shows notebook failure:
```
2022-03-24T21:34:18.3376638Z Notebook error:
2022-03-24T21:34:18.3377527Z AttributeError in src/Core/Transform/MutualInformationAffine/MutualInformationAffine.ipynb:
2022-03-24T21:34:18.3377846Z module 'jinja2.utils' has no attribute 'escape'
2022-03-24T21:34:18.3378212Z ninja: build stopped: subcommand failed.
2022-03-24T21:34:18.3378364Z ninja: build stopped: subcommand failed.
2022-03-24T21:34:18.3378495Z Command exited with the value: 1
2022-03-24T21:34:18.3381362Z Error(s) when building project
2022-03-24T21:34:18.3381875Z MakeCommand:/home/runner/work/_temp/685782585/cmake-3.18.3-Linux-x86_64/bin/cmake --build . --config "Release"
2022-03-24T21:34:18.3381982Z    0 Compiler errors
2022-03-24T21:34:18.3382098Z    1 Compiler warnings
2022-03-24T21:34:18.3382417Z SetCTestConfiguration:BuildDirectory:/home/runner/work/bld/ITKEx-build
```

CDash reports [1 compiler warning](https://open.cdash.org/viewBuildError.php?type=1&buildid=7814531) but seems to not report the trace above.

Unclear on the underlying cause as I'm not aware of any recent modifications to the notebook or build system.

tbirdso:
Looks like an issue in `jinja2` / `nbconvert`:

[https://github.com/jupyter/nbconvert/issues/1736](https://github.com/jupyter/nbconvert/issues/1736)

See that discussion for version workarounds.

tbirdso:
Looks like `jinja` issue is resolved, seeing what appears to be an issue with the Fleek API key for deployment:

```
Run fleekhq/action-deploy@v1
  with:
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.8.1[2](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/runs/5692386578?check_suite_focus=true#step:13:2)/x64
[...]

PR #336, opened 2022-03-25
BUG: Pin Jinja2 version

tbirdso:
Workaround for #335


dzenanz (approved):

tbirdso:
Logs show that the offending version of `jinja2` is still being installed after this change. It looks like the problem is under active investigation in `nbsphinx` which is likely installing `jinja2` as a dependency. In the interest of using our time effectively I am going to bump and revisit in a couple of days.

PR #338, opened 2022-03-29
WIP: Debug CI paths with tmate

tbirdso:
Adding CI step to examine state of Github Actions instance so that we can manually determine paths. Relevant to #337 

I believe I can get paths from the draft PR here.


tbirdso:
With tmate access into the runner I've verified that the directory does in fact exist:
```
runner@fv-az295-858:~/work$ find . -iname .fleek.json
./_actions/fleekhq/action-deploy/v1/.fleek.json
./Ex/.fleek.json
runner@fv-az295-858:~/work$ cd /home/runner/work/Ex
runner@fv-az295-858:~/work/Ex$
```

Perhaps `workDir` has to refer to a container path?

PR #340, opened 2022-03-30
BUG: Fix github actions variable ref

tbirdso:
I was a bit too trigger happy in #339 . Fixing CI variable.


tbirdso:
Intermediate CI results show that workspace variable is now properly referenced. Merging to get Fleek results.

PR #358, opened 2022-05-08
ENH: Turn off Git protocol by default.

andinet:


tbirdso (approved):
Thank you for fixing this @andinet !

Issue #362, opened 2022-05-16
Failure to find classes in `ITKDoxygenTAG/InsightDoxygen.tag`


tbirdso:
CI is failing with 6 warnings from missing TAG file definitions:

[CDash results](https://open.cdash.org/viewBuildError.php?type=1&buildid=7915737)

```none
2022-05-16T17:37:04.4831048Z /home/runner/work/bld/ITKEx-build/src/Filtering/ImageIntensity/AddConstantToEveryPixel/Documentation.rst:41: WARNING: doxygenclass: Cannot find class "itk::AddImageFilter" in doxygen xml output for project "ITK" from directory: /home/runner/work/bld/ITKEx-build/ITKDoxygenXML
2022-05-16T17:37:04.4832379Z /home/runner/work/bld/ITKEx-build/src/Filtering/ImageIntensity/AddConstantToEveryPixel/Documentation.rst:41: WARNING: Could not find match for `itk::AddImageFilter` in `/home/runner/work/bld/ITKEx-build/ITKDoxygenTAG/InsightDoxygen.tag` tag file. Error reported was Could not find a match
2022-05-16T17:37:04.4833848Z /home/runner/work/bld/ITKEx-build/src/Filtering/ImageIntensity/AddTwoImages/Documentation.rst:34: WARNING: doxygenclass: Cannot find class "itk::AddImageFilter" in doxygen xml output for project "ITK" from directory: /home/runner/work/bld/ITKEx-build/ITKDoxygenXML
2022-05-16T17:37:04.4875018Z /home/runner/work/bld/ITKEx-build/src/Filtering/ImageIntensity/AddTwoImages/Documentation.rst:34: WARNING: Could not find match for `itk::AddImageFilter` in `/home/runner/work/bld/ITKEx-build/ITKDoxygenTAG/InsightDoxygen.tag` tag file. Error reported was Could not find a match
2022-05-16T17:37:04.5222485Z /home/runner/work/bld/ITKEx-build/src/Numerics/Statistics/CreateHistogramFromListOfMeasurements/Documentation.rst:45: WARNING: doxygenclass: Cannot find class "itk::Statistics::SampleToHistogramFilter" in doxygen xml output for project "ITK" from directory: /home/runner/work/bld/ITKEx-build/ITKDoxygenXML
[...]

PR #365, opened 2022-05-18
BUG: Use Doxygen 5.2.0 archives rather than nightly

tbirdso:
Workaround to address #362 by getting Doxygen archives for ITK v5.2.0 rather than latest nightly.


thewtex (approved):
:+1: :green_heart: :1234:

PR #368, opened 2022-05-19
ENH: Add `ImageBufferAndIndexRange` example

N-Dekker:
Using `ImageBufferRange` and `IndexRange`, which were introduced with ITK 5.0.

### DIFF
--- a/src/Core/Common/ImageBufferAndIndexRange/Code.cxx
+++ b/src/Core/Common/ImageBufferAndIndexRange/Code.cxx
@@ -1,0 +1,87 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0.txt
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ *=========================================================================*/
+
+#include "itkImage.h"
+#include "itkImageBufferRange.h"
+#include "itkIndexRange.h"
+#include "itkNumericTraits.h"
+
+#include <cassert>
+#include <numeric> // For iota
+
+namespace
+{
+// Creates an image with sequentially increasing pixel values (0, 1, 2, ...).
+template <typename TImage>
+typename TImage::Pointer
+CreateImageWithSequentiallyIncreasingPixelValues(const typename TImage::RegionType & region)
+{
+  using PixelType = typename TImage::PixelType;
+
+  const auto image = TImage::New();
+  image->SetRegions(region);
+  image->Allocate();
+
+  const itk::ImageBufferRange<TImage> imageBufferRange(*image);
+  std::iota(imageBufferRange.begin(), imageBufferRange.end(), PixelType{ 0 });
+
+  return image;
+}
+
+
+// Prints all pixel values with their N-dimensional indices.
+template <typename TImage>
+void
+PrintPixelValues(const TImage & image)
+{
+  constexpr unsigned int                      Dimension{ TImage::ImageDimension };
+  const itk::ImageRegion<Dimension>           region = image.GetBufferedRegion();
+  const itk::ImageRegionIndexRange<Dimension> indexRange(region);
+  const itk::ImageBufferRange<const TImage>   imageBufferRange(image);
+
+  using PixelType = typename TImage::PixelType;
+  using PrintType = typename itk::NumericTraits<PixelType>::PrintType;
+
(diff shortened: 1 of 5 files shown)


src/Core/Common/ImageBufferAndIndexRange/Code.cxx
```diff
@@ -0,0 +1,78 @@
...
+CreateImageWithSequentiallyIncreasingPixelValues(const typename TImage::RegionType & region)
+{
+  using PixelType = typename TImage::PixelType;
+
+  const auto image = TImage::New();
+  image->SetRegions(region);
+  image->Allocate();
+
+  const itk::ImageBufferRange<TImage> imageBufferRange(*image);
+  std::iota(imageBufferRange.begin(), imageBufferRange.end(), PixelType{ 0 });
```
tbirdso, line 40: Very cool!

src/Core/Common/ImageBufferAndIndexRange/Code.cxx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+
+// Prints all pixel values with their N-dimensional indices.
+template <typename TImage>
+void
+PrintPixelValues(const TImage & image)
+{
+  const typename TImage::RegionType                        region = image.GetBufferedRegion();
+  const itk::ImageRegionIndexRange<TImage::ImageDimension> indexRange(region);
```
tbirdso, line 52: I am less familiar with `itk::ImageRegionIndexRange`, could you please comment on the advantage of this class over [`itk::ImageRegionIteratorWithIndex`](https://itk.org/Doxygen/html/classitk_1_1ImageRegionIteratorWithIndex.html)?

tbirdso (commented):
@N-Dekker Looking great! One question below.

tbirdso:
Two [CDash failures here](https://open.cdash.org/viewBuildError.php?type=1&buildid=7923770)

We may have difficulty in using classes in documentation that have been introduced after the 5.2 release, see [https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/issues/362](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/issues/362) discussion.

src/Core/Common/ImageBufferAndIndexRange/Code.cxx
N-Dekker: That's a very good question @tbirdso. Indeed, there is some overlap between the functionality of ImageRegionIndexRange and ImageRegionIteratorWithIndex. The most significant advantage of ImageRegionIndexRange is that it supports modern C++ iteration, including "range-based for loops". As included with the example:

    for (const auto & index : indexRange)
    {
      std::cout << "Index: " << index << '\n';
    }

Prints all the (N-dimensional) indices of pixels within the specified ImageRegion.

Another significant difference is that ImageRegionIndexRange is a more "lightweight" object than ImageRegionIteratorWithIndex. (Especially because `ImageRegionIndexRange` does not have a virtual function table, and it does not have a pointer to the image) So it might yield a better runtime performance. Especially for pixel-wise operations, this might be of interest.

Does that answer your question?

N-Dekker:
> Two [CDash failures here](https://open.cdash.org/viewBuildError.php?type=1&buildid=7923770)
> 
> We may have difficulty in using classes in documentation that have been introduced after the 5.2 release, see #362 discussion.

> ITKEx-build/src/Core/Common/ImageBufferAndIndexRange/Documentation.rst:57: WARNING: doxygenclass: Cannot find class "itk::ImageRegionIndexRange" in doxygen xml output for project "ITK" from directory: /.../bld/ITKEx-build/ITKDoxygenXML

"itk::ImageRegionIndexRange" is an alias, not a class. Maybe that's the problem...?

src/Core/Common/ImageBufferAndIndexRange/Code.cxx
tbirdso: Yes it does, very neat! Thank you 😄

thewtex (approved):
:man_singer: :sunglasses: awesome!!

N-Dekker:
Unfortunately "build-test-documentation (ubuntu-18.04)" still fails, even after my force-push. Any idea why?

N-Dekker:
FYI The [force-push](https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/compare/3979ef1b8c48276b5d6c1d56a95df30a28de1101..3a78d54217ed9a72861a1f946c0e3f463d902291) that I just did adjusted the `PrintPixelValues` example, because I did not like calling `image.GetPixel(index)` inside the `for` loop anymore. Because of course, `image.GetPixel(index)` is inefficient, when it is called for a whole range of consecutive pixels. While this is just example code, ITK users might simply copy the code for their applications and do such an iteration on larger images. So I thought, as an example, better use a range-based `for` loop over the image buffer range.  😃

thewtex (approved):
Awesome, Neils!

tbirdso:
🚀 Excellent example Niels! [https://examples.itk.org/src/core/common/imagebufferandindexrange/documentation](https://examples.itk.org/src/core/common/imagebufferandindexrange/documentation)

N-Dekker:
@tbirdso @thewtex Thanks for your encouragements and approvals. Still a very small improvement, adjusting the title of the example: pull request #386  😃

PR #369, opened 2022-05-19
STYLE: Remove parameters from `main(int, char *[])` when both are unused

N-Dekker:
I guess this is just a matter of taste. I just find those `int, char *[]` parameters distracting, whenever they are both unused.


thewtex:
Failed jobs restarted.

thewtex (approved):
:+1:

PR #370, opened 2022-05-19
STYLE: Remove extra spaces between parentheses of CMake command calls

N-Dekker:
Removed unnecessary extra space characters just after an opening, or just
before a closing parenthesis. It appeared more common _not_ to have those extra
space characters.

Did so by replacing `([a-z])\( ` with `$1\(`, and `(\S) \)` with `$1\)`, using
Notepad++ v8.3.3, Find in Files (Regular expression, Filters: `CMake*.txt`).


thewtex (approved):
:100: thanks!

PR #376, opened 2022-05-20
ENH: Adding python sample for ConvolveImageWithKernel

PranjalSahu:


src/Filtering/Convolution/ConvolveImageWithKernel/Code.py
```diff
@@ -0,0 +1,31 @@
...
+kernel = CreateSmoothKernel(width)
+kernel_image = itk.image_from_array(kernel)
+
+# Convolve image with kernel.
+filteredImage = itk.convolution_image_filter(inputImage, kernel_image=kernel_image)
+
+# Visualize the result using matplotlib
+import matplotlib.pyplot as plt
+plt.imshow(filteredImage)
+plt.show()
```
tbirdso, line 31: Per discussion, could you please add 1) output with `itk.imwrite` for baseline comparison against the C++ example, and 2) perhaps a screenshot of matplotlib output?

tbirdso (changes requested):

src/Filtering/Convolution/ConvolveImageWithKernel/Code.py
PranjalSahu: Added matplotlib visualization screenshot.

PranjalSahu: I will build it locally to see how it renders.

PranjalSahu:
It is failing due to automatically generated test for baseline comparison between C++ and Python
ConvolveImageWithKernelTestBaselineComparisonPython.

hjmjohnson:
Rebased onto current `main` in #453. Closing this in favor of the updated branch.

PR #394, opened 2022-08-19
Fix python3 find package

hjmjohnson:
Update to remove deprecated find_package(PythonLib) and find_package(PythonInterp) and the associated variables.

Use find_package(Python3) and its associated variables.


CMakeLists.txt
```diff
@@ -138,14 +138,21 @@ foreach(link ${content_links})
 endforeach()
 ExternalData_Add_Target(ITKSphinxExamplesData)
 
-if(${ITK_WRAP_PYTHON})
-  # Use the same Python interpreter as ITK to run the tests.
-  find_package(PythonInterp ${ITK_WRAP_PYTHON_VERSION} REQUIRED)
-  mark_as_advanced(PYTHON_EXECUTABLE)
+set(PYTHON_REQUIRED_VERSION 3.6)
+set(Python_ADDITIONAL_VERSIONS 3.12 3.11 3.10 3.9 3.8 3.7 3.6)
```
dzenanz, line 142: Python 3.6 reached end of life in December and is no longer supported. It should be removed from this list.

CMakeLists.txt
```diff
@@ -138,14 +138,21 @@ foreach(link ${content_links})
 endforeach()
 ExternalData_Add_Target(ITKSphinxExamplesData)
 
-if(${ITK_WRAP_PYTHON})
-  # Use the same Python interpreter as ITK to run the tests.
-  find_package(PythonInterp ${ITK_WRAP_PYTHON_VERSION} REQUIRED)
-  mark_as_advanced(PYTHON_EXECUTABLE)
+set(PYTHON_REQUIRED_VERSION 3.6)
```
dzenanz, line 141: This should be at least 3.7.

Superbuild/External-Python.cmake
```diff
@@ -1,12 +1,13 @@
-find_package(PythonInterp 3 REQUIRED)
+set(MIN_PYTHON_VERSION 3.5)
```
dzenanz, line 1: Also, 3.7 at least.

src/CMakeLists.txt
```diff
@@ -20,17 +20,11 @@ if(Module_ITKVtkGlue OR ITKVtkGlue_LOADED)
 endif()
 
 if(BUILD_DOCUMENTATION)
-  if(NOT PYTHON_EXECUTABLE)
-    find_package(PythonInterp 3.6 REQUIRED)
-    if(NOT PYTHONINTERP_FOUND)
+  set(PYTHON_REQUIRED_VERSION 3.6)
```
dzenanz, line 23: 3.7

dzenanz (commented):
Mostly looks good.

dzenanz (approved):

hjmjohnson:
@dzenanz Thanks for the comment.  I made this change, and also will be forcing ITK to 3.7 or greater soon.

thewtex:
@hjmjohnson could this please be rebased on `master`?

jhlegarreta:
> @dzenanz Thanks for the comment. I made this change, and also will be forcing ITK to 3.7 or greater soon.

The CI workflow already uses 3.9,
https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/blob/master/.github/workflows/build-test-publish.yml#L37

and Python 3.7 is 5 months from deprecation, so if a minimum version needs to be enforced, maybe it should be 3.8. At the same time:
- We are 12 months away from 3.8 deprecation
- I have also the feeling that many have jumped to 3.10 from 3.8, so not sure if 3.9 has been a widely adopted version.

hjmjohnson:
This works locally, but the GitHub actions are all broken for all PRS.

PR #398, opened 2022-11-07
ENH: Migrate data to IPFS

thewtex:

PR #399, opened 2022-11-10
ENH: Update CMake to decouple archives from example builds

### LINKED ISSUE #373: BUG: Example zip archive downloads are absent
ZIP archives are absent from the server.

[https://examples.itk.org/src/filtering/convolution/colornormalizedcorrelation/ColorNormalizedCorrelation.zip](https://examples.itk.org/src/filtering/convolution/colornormalizedcorrelation/ColorNormalizedCorrelation.zip)

tbirdso:
Addresses issue where archives were not build for examples without an `add_example()` call, resulting in broken download links on `examples.itk.org`.

New behavior packages example archives for all examples to accompany Sphinx documentation. Also cleans up `CMakeLists.txt` for several module groups.

Closes #373.

### DIFF
--- a/src/Filtering/CMakeLists.txt
+++ b/src/Filtering/CMakeLists.txt
@@ -1,26 +1,39 @@
-add_subdirectory_if_module_enabled(BinaryMathematicalMorphology)
-add_subdirectory_if_module_enabled(ImageGrid)
-add_subdirectory_if_module_enabled(ImageFeature)
-add_subdirectory_if_module_enabled(FastMarching)
-add_subdirectory_if_module_enabled(Colormap)
-add_subdirectory_if_module_enabled(ImageIntensity)
-add_subdirectory_if_module_enabled(ImageFilterBase)
-add_subdirectory_if_module_enabled(Thresholding)
-add_subdirectory_if_module_enabled(QuadEdgeMeshFiltering)
-add_subdirectory_if_module_enabled(ImageCompare)
-add_subdirectory_if_module_enabled(MathematicalMorphology)
-add_subdirectory_if_module_enabled(Smoothing)
-# TODO: Need to re-enable ImageGraadient
-# add_subdirectory_if_module_enabled(ImageGradient)
-add_subdirectory_if_module_enabled(AnisotropicSmoothing)
-add_subdirectory_if_module_enabled(FFT)
-add_subdirectory_if_module_enabled(AntiAlias)
-add_subdirectory_if_module_enabled(LabelMap)
-add_subdirectory_if_module_enabled(ImageStatistics)
-add_subdirectory_if_module_enabled(ImageFusion)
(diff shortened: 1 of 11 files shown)


dzenanz (approved):
I have not tested anything here. Code looks good.

tbirdso:
Seeing repeated failures due to SHA256 resource download failures, which are in the process of being addressed with https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/pull/398.

https://open.cdash.org/viewBuildError.php?buildid=8270992

Documentation builds on my Ubuntu machine and I have verified that archives are generated based on folder structure to address #373. @thewtex Would you prefer that I wait and rebase once #398 is stable or go ahead and merge?

src/Filtering/CMakeLists.txt
```diff
@@ -1,26 +1,34 @@
...
+  ImageStatistics
+  LabelMap
+  MathematicalMorphology
+  "Path"
+  QuadEdgeMeshFiltering
+  Smoothing
+  Thresholding
+)
+
+foreach(_module ${Filtering_MODULES})
```
thewtex, line 29: A comment here would be helpful to explain what ImageGradient is not enabled

src/Filtering/CMakeLists.txt
```diff
@@ -1,26 +1,34 @@
...
+  ImageFilterBase
+  ImageFusion
+  ImageGradient
+  ImageGrid
+  ImageIntensity
+  ImageLabel
+  ImageStatistics
+  LabelMap
+  MathematicalMorphology
+  "Path"
```
thewtex, line 23: Why is this quoted?

thewtex (approved):
Great, thanks @tbirdso !

We can merge if we can get a green build

src/Filtering/CMakeLists.txt
tbirdso: Quotes are to distinguish to the reader that we are using "Path" as a module name rather than as the CMake `PATH` keyword or the system `PATH` variable. Without the quotes my IDE highlights `Path` as a keyword. Would it be less confusing to remove the quotes?

thewtex: Yes, it would be helpful to remove the quotes or add a comment explaining why they are there.

src/Filtering/CMakeLists.txt
tbirdso: At least one ImageGradient example is unstable, I received a compiler error when I attempted to re-enable it. It looks like ImageGradient examples have been disabled since a large addition back in 2018. There is more work to be done in investigating and addressing those issues, but it is outside the scope of this PR.

src/Filtering/CMakeLists.txt
tbirdso: Unquoted

PR #412, opened 2023-01-10
STYLE: Capitalize first letter for `Histogram` class name in index

jhlegarreta:
Capitalize first letter for `Histogram` class name in index: avoids
creating a `histogram` entry linked to the same class name, and adds a
new link in the `Histogram` index entry.

### DIFF
--- a/src/Numerics/Statistics/CreateHistogramFromListOfMeasurements/Documentation.rst
+++ b/src/Numerics/Statistics/CreateHistogramFromListOfMeasurements/Documentation.rst
@@ -5,7 +5,7 @@ Create Histogram From List of Measurements
 
 .. index::
    single: SampleToHistogramFilter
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

jhlegarreta:
The use of `cast` _might_ be OK:
https://examples.itk.org/genindex.html

If we decide that it may be used to group all examples performing some casing. Maybe it should be kept and additionally add `CastImageFilter` to the index. What do you think?

dzenanz:
I don't have much of an opinion there. Maybe Matt and Tom do?

jhlegarreta:
So it turns out that this is not really a bug, and it is more of a style thing/creating an entry that better identifies the class name.

dzenanz (approved):

PR #416, opened 2023-02-09
ENH: Add Python version of ExtractIsoSurface

thewtex:


src/Core/Mesh/ExtractIsoSurface/CMakeLists.txt
```diff
@@ -13,16 +13,31 @@ install(TARGETS ${PROJECT_NAME}
...
-install(FILES Code.cxx CMakeLists.txt
+install(FILES Code.cxx CMakeLists.txt Code.py
   DESTINATION share/ITKSphinxExamples/Code/Core/Mesh/ExtractIsoSurface
   COMPONENT Code
   )
 
+set(input_volume ${CMAKE_CURRENT_BINARY_DIR}/HeadMRVolume.mha)
+set(output_mesh ${CMAKE_CURRENT_BINARY_DIR}/Output.vtk)
+set(lower_threshold 64)
+set(upper_threshold 255)
```
tbirdso, line 24: Nitpick, could these please each be followed by an `unset` after their use to avoid accidentally propagating to other examples?

src/Core/Mesh/ExtractIsoSurface/Code.py
```diff
@@ -0,0 +1,43 @@
...
+input_image = itk.imread(args.input_image)
+
+thresholded = itk.binary_threshold_image_filter(
+    input_image,
+    lower_threshold=args.lower_threshold,
+    upper_threshold=args.upper_threshold,
+    outside_value=0,
+)
+
+output_mesh = itk.binary_mask3_d_mesh_source(thresholded, object_value=255)
```
tbirdso, line 41: The default value for [itkBinaryThresholdImageFilter::m_InsideValue](https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/Filtering/Thresholding/include/itkBinaryThresholdImageFilter.h#L75) is the max for the given scalar datatype, so as I understand it `255` would not be valid for pixel data types larger than `itk.UC`. An alternative would be to specify `inside_value=255` in calling `itk.binary_threshold_image_filter` above.

tbirdso (approved):
Looks good! Proposed a couple of optional changes.

src/Core/Mesh/ExtractIsoSurface/CMakeLists.txt
thewtex: I understand the intention to use unset to avoid interference with other examples, however, this content would be a major distraction in the CMakeLists.txt for those trying to learn from the individual example -- it is not related to the individual sample. This is much more important. Also, are you sure that this is actually going to cause issues with other examples due to CMake scoping via `add_subdirectory`?

src/Core/Mesh/ExtractIsoSurface/Code.py
thewtex: :+1: `inside_value=255` was added to `itk.binary_threshold_image_filter`.

src/Core/Mesh/ExtractIsoSurface/CMakeLists.txt
tbirdso: Ah, you are correct, scoping means this is not an issue. Good to leave as-is.

src/Core/Mesh/ExtractIsoSurface/Code.py
tbirdso: Thank you!

tbirdso (approved):

PR #425, opened 2024-06-02
COMP: Disable nbsphinx

thewtex:
The sphinx build is hanging when nbsphinx is parsing the notebooks with
the recent sphinx/nbsphinx stack. Disable nbsphinx for now.


hjmjohnson:
Closing as stale. This workaround to disable `nbsphinx` was opened 2 years ago to address a hanging Sphinx build, but has not been updated. The `build-test-documentation` CI failure is being tracked separately (see #448 PR description) as part of a broader documentation CI fix effort. If nbsphinx needs to be disabled, that should be addressed in that broader context rather than via this stale branch.

Thank you @thewtex for identifying and reporting the issue.

PR #452, opened 2026-04-17
ENH: Migrate example itkGrayscaleFunctionDilateImageFilter

hjmjohnson:
Rebase of @jadh4v's PR #374 onto current `main`.

Migrates the `itkGrayscaleFunctionDilateImageFilter.cxx` example from `ITK/Examples/Filtering` to `ITKSphinxExamples/src/Filtering/MathematicalMorphology/DilateUsingFunctionalGrayscale`. Adds C++, Python, documentation, and a new baseline output image (`OutputBaseline.png`).

PR #453, opened 2026-04-17
ENH: Add Python example for ConvolveImageWithKernel

hjmjohnson:
Rebase of @PranjalSahu's PR #376 onto current `main`.

Adds a Python (`Code.py`) example for `ConvolveImageWithKernel`, complementing the existing C++ example.


src/Filtering/Convolution/ConvolveImageWithKernel/MatplotlibVisualizeConvolution.png.sha512
```diff
@@ -0,0 +1 @@
+082d1d49439879f93946e39675c62a6732fbb1d17aeb7ecbc63c6ca7e350b34416d3764c355e4c1c00583fd8231eb79f01cc6c618d5cda7eee8a267df64db7a3
```
dzenanz: CID?

dzenanz (commented):
Mostly looks good.

src/Filtering/Convolution/ConvolveImageWithKernel/MatplotlibVisualizeConvolution.png.sha512
hjmjohnson: Good catch — agreed it should move to a `.cid` link. The content isn't on IPFS yet (only in the legacy sha512 store), so I can't flip the extension without breaking ExternalData resolution. The PNG (sha512 `082d1d49…`) computes to CID `bafkreiejmu6wjzs35rezwhe4htcehxslunbdju424td6zjy3giucxgczvu`; it needs to be published to ITKTestingData first (same dependency as the cthead1/OutputBaseline fixtures in #452 → ITKTestingData#67). Will convert once it's pinned. The current `.sha512` link still resolves, so CI is otherwise unaffected.

hjmjohnson: Done in 242b5d1e — converted to `MatplotlibVisualizeConvolution.png.cid` (`bafkreiejmu6wjzs35rezwhe4htcehxslunbdju424td6zjy3giucxgczvu`). The blob is now published in ITKTestingData via InsightSoftwareConsortium/ITKTestingData#68 (merged).

PR #467, opened 2026-06-15
Use FetchContent to build ITK

blowekamp:
Remove the External ITK Superbuild, and replace with a tightly coupled ITKFetchContent.


dzenanz (approved):
Looks good on a glance.

CMake/ITKFetchContent.cmake
```diff
@@ -0,0 +1,60 @@
+#-----------------------------------------------------------------------------
```
dzenanz, line 1: This file addition should be in a different commit.

CMake/ITKFetchContent.cmake
```diff
@@ -0,0 +1,60 @@
...
+endif()
+
+# Configure ITK-specific options that need to be set before FetchContent_Declare
+if(NOT ${BUILD_SHARED_LIBS})
+  set(CMAKE_C_VISIBILITY_PRESET "hidden")
+  set(CMAKE_CXX_VISIBILITY_PRESET "hidden")
+  set(ITK_TEMPLATE_VISIBILITY_DEFAULT OFF)
+endif()
+
+set(Modile_ITKReview ON CACHE INTERNAL "")
```
hjmjohnson: Modile_ITKReview -> Module_ITKReview

CMake/ITKFetchContent.cmake
```diff
@@ -0,0 +1,59 @@
...
+# Set ITK Git repository and tag
+set(ITK_GIT_REPOSITORY "https://github.com/InsightSoftwareConsortium/ITK.git")
+set(ITK_GIT_TAG "main")
+
+if(NOT DEFINED ITK_BUILD_DEFAULT_MODULES)
+  set(ITK_BUILD_DEFAULT_MODULES ON)
+endif()
+
+# Configure ITK-specific options that need to be set before FetchContent_Declare
+if(NOT ${BUILD_SHARED_LIBS})
```
hjmjohnson: if(NOT ${BUILD_SHARED_LIBS}) is fragile. If BUILD_SHARED_LIBS is undefined the line expands to if(NOT ),  evaluating TRUE and forcing hidden visibility + ITK_TEMPLATE_VISIBILITY_DEFAULT OFF (static-style). Use if(NOT BUILD_SHARED_LIBS) (unquoted variable name) so it reads the value, not a dereferenced empty.

hjmjohnson (changes requested):
Just a few minor fixes.

CMake/ITKFetchContent.cmake
blowekamp: done

CMake/ITKFetchContent.cmake
blowekamp: done

hjmjohnson (approved):

dzenanz (approved):

thewtex (approved):
👍

PR #469, opened 2026-06-16
Enable and format with gersemi for CMake style

blowekamp:


dzenanz:
Could you split this into two commits, one doing infrastructure changes, and the second applying the style to CMake code?

blowekamp:
> Could you split this into two commits, one doing infrastructure changes, and the second applying the style to CMake code?

done

dzenanz (approved):

PR #470, opened 2026-06-16
WIP: Use interface library variable

blowekamp:
This is currently not working as I expected. The ITK_INTERFACE_LIBRARIES contains all the ITK modules, but not the meta-modules to enable the factories.


blowekamp:
One underlying problem with this PR is that multiple calls to find_package are caches so the components of the first one are return in subsequent calls.

PR #476, opened 2026-09-02
COMP: Keep GeodesicActiveContour example working on ITK 5.4 and 6.0

hjmjohnson:
`SegmentWithGeodesicActiveContourLevelSet/Code.py` hard-codes `itk.VectorContainer[itk.UI, LevelSetNode]`, which ITK 6.0 no longer wraps after InsightSoftwareConsortium/ITK#6803. Selecting the index type at runtime keeps the example working on both 5.4 and 6.0.

> Stacked on #475 — that PR fixes an unrelated `pre-commit` deadlock that blocks `--all-files` on `main`. Review only the second commit here; this becomes a one-commit PR once #475 merges.

<details>
<summary>Why a single spelling cannot cover both</summary>

ITK#6803 changes `LevelSetTypeDefault::NodeContainer` from `VectorContainer<unsigned int, NodeType>` to the one-argument `VectorContainer<NodeType>`, whose index type is `SizeValueType`, and updates `ITKFastMarchingBase.wrap` accordingly.

The wrapping registry is keyed by the mangled name emitted by that one `itk_wrap_template` call, so the two spellings are strictly disjoint across the version boundary — there is no release where both resolve:

| ITK | `itk.VectorContainer[itk.UI, LevelSetNode]` | `itk.VectorContainer[itk.IT, LevelSetNode]` |
|---|---|---|
| 5.4.x | resolves | `TemplateTypeError` |
| 6.0 | `TemplateTypeError` | resolves |

Confirmed empirically against ITK 5.4.6, and `ITKFastMarchingBase.wrap` is byte-identical at `v5.4.7`, `release-5.4` and `main`-before-#6803. `itk.TemplateTypeError` subclasses `TypeError`, which is what the guard catches.

</details>

<details>
<summary>Test</summary>

Ran the example end to end against ITK 5.4.6 on `BrainProtonDensitySlice.png` (the `except` branch, which ITK's own CI cannot reach): exit 0, 171 iterations, RMS change 0.0198, output written.

`NodeType` is also hoisted so the template is spelled once rather than twice.

</details>

### DIFF
--- a/src/External/CMakeLists.txt
+++ b/src/External/CMakeLists.txt
@@ -1,1 +1,2 @@
-
+# No examples in this category yet; this file exists because
+# src/CMakeLists.txt calls add_subdirectory(External).
--- a/src/Filtering/AnisotropicSmoothing/ComputeCurvatureAnisotropicDiffusion/CMakeLists.txt
+++ b/src/Filtering/AnisotropicSmoothing/ComputeCurvatureAnisotropicDiffusion/CMakeLists.txt
@@ -2,7 +2,12 @@ cmake_minimum_required(VERSION 3.22.1)
 
 set(input_image ${CMAKE_CURRENT_BINARY_DIR}/BrainProtonDensitySlice.png)
 set(output_image Output.png)
-set(test_options 5 0.125 3.0)
+set(
+  test_options
+  5
+  0.125
+  3.0
+)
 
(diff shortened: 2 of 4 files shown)


dzenanz (commented):
Looks good. Rebase on main before merging.

PR #481, opened 2026-09-12
WIP: Require ITK 6 and request modules as find_package COMPONENTS

hjmjohnson:
Request the ITK modules each example uses as `find_package` COMPONENTS.
Link lines are unchanged: each example keeps its explicit
`ITK::<Name>Module` targets.

Stacked on #479 and #482 — review the last 10 commits.

<details>
<summary>What this buys</summary>

Components are derived from the ITK headers each example includes, resolved through an index of 3282 headers across ITK's 167 modules. Naming them means a missing or disabled module fails at configure time with `No such module: X`, rather than as a missing target at generate time.

Link lines keep the explicit module targets rather than `${ITK_INTERFACE_LIBRARIES}`: these examples are teaching material, and `ITK::ITKThresholdingModule` on the link line tells a reader which module provides the class the example demonstrates.

</details>

<details>
<summary>Why the components are not narrowed to specific IO formats</summary>

Requesting the exact format an example reads (`ITKIOPNG` instead of the `ITKImageIO` meta-module) shrinks the built examples from 5.26 GB to 0.72 GB, because only the requested factory is registered.

That is not usable here yet: `itk_generate_factory_registration()` records the generated header's include directory on the shared `ITK::ITKImageIO` target, so in an aggregate build the last example configured wins and the others compile against its registration header. Measured on this repository, per-format narrowing produced 62 new test failures for that reason. Reported as InsightSoftwareConsortium/ITK#6860; once registration attaches per target, the narrowing becomes available.

Six examples still name a format module — the GDCM, TIFF, and factory-registration examples — because they include `itkGDCMImageIO.h`, `itkTIFFImageIO.h`, or `itkPNGImageIOFactory.h` directly and need it on the include path.

</details>

<details>

Built against ITK `main` with `Module_ITKVtkGlue=ON` and Python wrapping enabled:

| | before | after |
|---|---|---|
| build | 0 failed targets, 365 executables | 0 failed targets, 365 executables |
| tests | 2 failures of 558 | 2 failures of 558 |

No new failures and none fixed. The two remaining failures are unrelated to this change: `ConvolveImageWithKernelTestPython` needs `matplotlib`, and `ComputeFFTInOneDimensionPhaseBaselineComparisonPython` fails a baseline image comparison.

</details>


src/Filtering/AnisotropicSmoothing/ComputeCurvatureAnisotropicDiffusion/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ add_executable(${PROJECT_NAME} Code.cxx)
 target_link_libraries(
   ${PROJECT_NAME}
   PRIVATE
+    ${ITK_INTERFACE_LIBRARIES}
     ${ITK_INTERFACE_LIBRARIES}
```
hjmjohnson, line 29: Duplicates here

src/Filtering/AnisotropicSmoothing/ComputeCurvatureFlow/CMakeLists.txt
```diff
@@ -10,16 +10,22 @@ set(
...
+    ITKImageIntensity
+)
 itk_generate_factory_registration()
 
 add_executable(${PROJECT_NAME} Code.cxx)
 target_link_libraries(
   ${PROJECT_NAME}
   PRIVATE
+    ${ITK_INTERFACE_LIBRARIES}
     ITK::ITKImageIO
```
hjmjohnson, line 28: Why does ITK::ITKImageIO survive here?

src/Filtering/AnisotropicSmoothing/ComputeCurvatureAnisotropicDiffusion/CMakeLists.txt
hjmjohnson: Fixed in 12e0c31a. This file had already been converted by the reference-example commit, and the Filtering pass then inserted a second `${ITK_INTERFACE_LIBRARIES}` — the conversion script was not idempotent. A repo-wide scan confirms this was the only file affected.

src/Filtering/AnisotropicSmoothing/ComputeCurvatureFlow/CMakeLists.txt
hjmjohnson: Required, not residue. Linking the factory meta-module is what enables the generated registration code, so it stays even though the module is also a requested component.

Checked rather than assumed: building this example without `ITK::ITKImageIO` still compiles, but the binary aborts at runtime (SIGABRT, no output written) because no ImageIO factory is registered. With it, the example runs and writes its output.

blowekamp:
What was the prior behavior here for these examples? Was every example just linked to everything in ITK and all ImageIO's loaded?

It appears a motivation for using only the needed ImageIO is to reduce the binary size. I would be interested in how the size of a share library build is by comparison to the other options.

I think finding the "ImageIO" package is the best approach, which is now done here.

dzenanz:
> Was every example just linked to everything in ITK and all ImageIO's loaded?

This is quite possible. The examples' build directory was big: 30-40 GB with statically built libraries.

src/Bridge/VtkGlue/ConvertvtkImageDataToAnitkImage/CMakeLists.txt
dzenanz, line 9: We no longer support VTK 8.x, so this branch can be removed. The same applies to many other files in Bridge group.

dzenanz (commented):
Mostly looks good

src/Bridge/VtkGlue/ConvertvtkImageDataToAnitkImage/CMakeLists.txt
hjmjohnson: https://github.com/InsightSoftwareConsortium/ITKSphinxExamples/pull/488

hjmjohnson:
Closing: once #479, #482, and #488 are in, requesting modules as COMPONENTS shows no measured benefit, so there is nothing left worth merging here.

> What was the prior behavior here for these examples? Was every example just linked to everything in ITK and all ImageIO's loaded?

All ImageIOs, not all of ITK. Each example links explicit `ITK::<Name>Module` targets, and 247 of them include `ITK::ITKImageIO`, which registers every enabled image IO factory. A bare `find_package(ITK REQUIRED)` loads every module's config but links none of them.

> I would be interested in how the size of a share library build is by comparison to the other options.

With `ITKImageIO` requested, COMPONENTS changes nothing in the output. I agree that requesting the `ITKImageIO` package is the right choice.

<details>
<summary>Measurements (static ITK main, macOS arm64)</summary>

One standalone example (`MedianFilteringOfAnImage`), bare `find_package(ITK REQUIRED)` compared with `COMPONENTS ITKCommon ITKSmoothing ITKImageIO`:

| | bare | COMPONENTS |
|---|---|---|
| configure time (warm) | 3.96 s | 3.90 s |
| image IO factories registered | 24 | 24 |
| binary size | 23.73 MB | 23.73 MB |
| missing module | configure error | configure error |

Across the aggregate build, `bin/` is 5.42 GB for 372 executables with or without COMPONENTS.

The only size win is narrowing to specific formats (e.g. `ITKIOPNG`). That cut the built examples from 5.26 GB to 0.72 GB, but it produced 62 test failures. The cause is ITK#6860: the factory registration include directory is set on the shared `ITK::ITKImageIO` target, so the last example configured wins. Revisit once that is fixed.

A shared-library comparison has not been measured.

</details>

PR #482, opened 2026-09-12
COMP: Require ITK 6

hjmjohnson:
Raise the minimum ITK to 6 in `src/CMakeLists.txt`. Every example already
links `ITK::<Name>Module` interface targets, which ITK 5.x does not export.

<details>
<summary>What ITK 5.x does today, and what this changes</summary>

Against ITK 5.4 the build currently fails with one missing-target error per
example — 346 in all — because the namespaced interface targets were introduced
in ITK 6. Requiring ITK 6 at the top level stops at the first `find_package`
with a single version error instead:

```
CMake Error at src/CMakeLists.txt:7 (find_package):
  Could not find a configuration file for package "ITK" that is compatible
  with requested version "6".
```

Three examples still named an older minimum (5.0, 5.0, 4.7.0) from before the
CMake refactoring. They now use a bare `find_package(ITK REQUIRED)` like the
other 408 examples, so the requirement is stated once, in `src/CMakeLists.txt`.

</details>

<details>
<summary>Where ITK 5.x examples live</summary>

The `release-5.4` branch was created at `b0679d89`, the last commit that still
builds against ITK 5.x. Verified against ITK 5.4.7: 0 failed targets, 346
executables, 370 of 370 tests passing.

</details>


src/Core/Common/FilterAndParallelizeImageRegion/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.22.1)
 
 project(FilterAndParallelizeImageRegion)
 
-find_package(ITK 5.0 REQUIRED)
```
dzenanz, line 5: This appears in a lot more places than just these 4.

dzenanz (changes requested):

src/Core/Common/FilterAndParallelizeImageRegion/CMakeLists.txt
dzenanz: These 4 example required ITK 5 one upon a time, but with CMake refactoring, all the modules now require ITK 6.

hjmjohnson: Agreed — the per-example version pins were leftovers from before the CMake refactoring. Fixed in 314ed1d7: these three now use a bare `find_package(ITK REQUIRED)` like the other 408 examples, and the ITK 6 requirement is stated once, in `src/CMakeLists.txt`.

Two other remnants of old version checks turned up while looking, and are left for separate PRs: ten Python examples still guard against ITK 4.7–5.2 via `GetITKVersion()`, and `src/Bridge/VtkGlue/CMakeLists.txt` gates seven examples behind `if(ITK_VERSION_MAJOR LESS 5)`.

dzenanz (approved):

PR #485, opened 2026-09-14
ENH: Restore the static LevelSetsv4 examples without VTK rendering

hjmjohnson:
Restore five LevelSetsv4 examples excluded since ITK 5 by writing their output
instead of rendering through VTK classes ITK no longer has. They move to
`Segmentation/LevelSetsv4`.

<details>
<summary>Why these examples, and why they move</summary>

These are the only examples of the LevelSetsv4 framework — `LevelSetDenseImage`,
the Whitaker, Shi, and Malcolm sparse level sets, and
`BinaryImageToLevelSetImageAdaptor` appear in no other example. They were built
only under `if(ITK_VERSION_MAJOR LESS 5)` because their visualization classes
moved out of ITK in 2019 (`LevelSetsv4Visualization`, now in the separate
ITKVtkGlue repository). The level-set code itself is unchanged ITK 6 API.

Once they no longer render, nothing ties them to VTK, so they move from
`Bridge/VtkGlue` to `Segmentation/LevelSetsv4`, matching the ITK module.

</details>

<details>
<summary>What each example now writes</summary>

The two dense examples used to differ only in how they rendered, so each now
writes what its name describes:

| example | output |
|---|---|
| `VisualizeStaticDense2DLevelSetAsElevationMap` | dense level-set values |
| `VisualizeStaticDense2DLevelSetZeroSet` | zero set, via `ZeroCrossingImageFilter` |
| `VisualizeStatic{Whitaker,Shi,Malcolm}2DLevelSetLayers` | sparse layers, sampled on the image grid |

All five also fix a latent bug: the Otsu mask pipeline was never updated before
the level set was initialized, so the level set started from an empty mask and
every pixel landed outside it.

The rendered screenshot (`levelsets.png`) is removed from each page, since it
showed a renderer that no longer exists.

</details>

<details>

Built against ITK `main`: 0 build failures, and all 5 new tests pass. Each
output was inspected, not just checked for existence:

| example | output |
|---|---|
| Dense ElevationMap | values −10.2 to 42.3, 2147 pixels inside |
| Dense ZeroSet | 1914-pixel contour |
| Whitaker | layers −3, −2, −1, 0, 1, 2, 3 |
| Shi | layers −3, −1, 1, 3 |
| Malcolm | layers −1, 0, 1 |

Commit 1 is a pure move (25 renames at 100% similarity), so history follows the
files. `pre-commit run --all-files` passes.

The tests check that each example runs; they do not compare against a baseline
image, which would first need new test data published.

</details>


dzenanz (approved):
Looks good on a glance.

dzenanz:
Some of Kitware's services are currently under DDoS attack. [cmake.org](cmake.org) timeout is real, but temporary.

PR #486, opened 2026-09-14
ENH: Restore the evolving LevelSetsv4 examples without VTK rendering

hjmjohnson:
Restore the two evolving LevelSetsv4 examples by writing the evolved level set
instead of rendering each iteration through VTK, and remove the now-empty
`if(ITK_VERSION_MAJOR LESS 5)` block.

<details>
<summary>What changes</summary>

Both examples evolve a dense level set with the Chan and Vese region terms. They
used to attach `LevelSetIterationUpdateCommand` to a VTK visualizer that ITK no
longer contains; each now writes the result after evolving:

| example | output |
|---|---|
| `VisualizeEvolvingDense2DLevelSetAsElevationMap` | evolved level-set values |
| `VisualizeEvolvingDense2DLevelSetZeroSet` | evolved zero set, via `ZeroCrossingImageFilter` |

They move to `Segmentation/LevelSetsv4` alongside the static examples added in #485.
With them gone, the VtkGlue `if(ITK_VERSION_MAJOR LESS 5)` block is empty and is
removed. The screenshot and animation (`levelsets.png`, `levelsets.gif`) are
removed from each page, since they showed a renderer that no longer exists.

</details>

<details>

Built against ITK `main`: 0 build failures, and all 7 LevelSetsv4 tests pass
(the 5 from #485 plus these 2, each evolving 100 iterations). Outputs inspected:

| example | output |
|---|---|
| Evolving ElevationMap | values −13.0 to 42.3, 2816 pixels inside |
| Evolving ZeroSet | 1627-pixel contour |

Commit 1 is a pure move (12 renames), so history follows the files.
`pre-commit run --all-files` passes. As in #485, the tests check that each
example runs rather than comparing against a baseline image.

</details>


dzenanz (approved):
Looks good on a glance.

dzenanz (approved):

